PR #24, opened 2015-06-08
Minor fixes for actions guides

tomkadwill:


jodosha:
@tomkadwill Thank you :panda_face:

PR #38, opened 2015-06-23
Fix Avdi's name

jc00ke:


jodosha:
@jc00ke @avdi ouch :wink: Thank you.

PR #40, opened 2015-06-23
Correct file system adapter connection string

thegcat:


jodosha:
@thegcat Thank you! :+1:

PR #41, opened 2015-06-24
Fix small amount of markup and spelling in views/templates.md

djfpaagman:
Fixed some small stuff I found while reading the guide


jodosha:
@djfpaagman Thank you very much! :+1:

PR #42, opened 2015-06-24
Remove view partials from navigation and fix link to database cmd line tools

djfpaagman:
1. I removed the link to the view partials guide, as it does not seem to exist. Correct me if i'm wrong :)
2. Fixed a link to the database command line guide in the Migration Overview article


jodosha:
@djfpaagman Thank you again.

PR #43, opened 2015-06-24
Ruby uses native Threads

tak1n:
http://www.csinaction.com/2014/10/10/multithreading-in-the-mri-ruby-interpreter/


jodosha:
@tak1n ooops, thank you. :+1:

PR #45, opened 2015-06-25
Fix null-constraint violation exception in getting started guides

Bounga:
books table migration adds two columns (created_at, updated_at) which
have a null-constraint. To be able to persist `Book` objects we have
to add `created_at` and `updated_at` to `Book` entity attributes and
map them accordingly.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -407,8 +407,6 @@ Lotus::Model.migration do
       primary_key :id
       column :title,      String,   null: false
       column :author,     String,   null: false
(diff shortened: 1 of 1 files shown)


jodosha:
@Bounga I edited that article, my intention was to remove `created_at`/`updated_at` references to keep it simple.

Would you mind to amend this PR and remove all their occurrences? Thank you :)

Bounga:
I'll do this!

Bounga:
@jodosha done.

jodosha:
@Bounga Thank you, merging it! :shipit:

PR #46, opened 2015-06-25
Fix books controller specs for exposure in getting started guide

Bounga:
According to the guide, this action purpose is to expose all books
to the view. It now compares exposed books to those available in
`BookRepository` rather than an empty array.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -572,6 +572,12 @@ describe Web::Controllers::Books::Index do
   let(:action) { Web::Controllers::Books::Index.new }
   let(:params) { Hash[] }
 
+  before do
+    BookRepository.clear
+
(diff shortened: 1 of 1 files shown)


source/guides/getting-started.md
```diff
@@ -579,7 +579,7 @@ describe Web::Controllers::Books::Index do
 
   it 'exposes all books' do
     action.call(params)
-    action.exposures[:books].must_equal []
+    action.exposures[:books].must_equal BookRepository.all
```
jodosha: @Bounga This requires a clean database table before the test starts. I'd go for a more explicit way, like the hardcoded array. Example: `[]` or `[book]` where `book` is an entity that we have persisted in the test setup.

Bounga:
I updated this PR to match your comment. Is it more in the spirit you were talking about?

jodosha:
@Bounga Perfect :+1:

PR #47, opened 2015-06-25
Fix book create action specs in getting started guide

Bounga:
- add missing `action` variable
- cleanup DB before starting tests
- use `BookRepository` to check if a record was created since calling
`action.book` won't work. `book` isn't an instance variable and isn't
exposed by the controller and doesn't have to be.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -713,15 +713,16 @@ require 'spec_helper'
 require_relative '../../../../apps/web/controllers/books/create'
 
 describe Web::Controllers::Books::Create do
+  let(:action) { Web::Controllers::Books::Create.new }
   let(:params) { Hash[book: { title: 'Confident Ruby', author: 'Avdi Grimm' }] }
 
-  after do
+  before do
     BookRepository.clear
   end
 
   it 'creates a new book' do
     action.call(params)
-    action.book.id.wont_be_nil
+    BookRepository.all.count.wont_equal 0
   end
 
   it 'redirects the user to the books listing' do


source/guides/getting-started.md
```diff
@@ -713,15 +713,16 @@ require 'spec_helper'
...
 
-  after do
+  before do
     BookRepository.clear
   end
 
   it 'creates a new book' do
     action.call(params)
-    action.book.id.wont_be_nil
+    BookRepository.all.count.wont_equal 0
```
jodosha: @Bounga I don't think this is a good idea, because it references `BookRepository` directly.

If you change data store, or want to use a stub repository to inject for this test, you'll be in trouble. It also requires that your database table **must** be empty before the test, otherwise you'll have a false positive pass.

Exposures are a great way to unit test actions, their usage should be encouraged.

Can you please remove this change, and I'll merge it. Thank you! :smile:

Bounga: So I have to expose `book` variable then use it in the spec?

Is it recommended to modify action code only to please specs?

jodosha: Good point, the book isn't gonna be used from any view, because we're doing a redirect. So there isn't a need to expose it.

Bounga: So I don't really know how to avoid `BookRepository` call in this test. What would be the "lotus" way? Or at least the cleanest OOP way of doing this?

joneslee85: I don't think there is a "lotus" way for doing this, maybe it's better to extract the logic to BookCreation interactor class and test that controller has called that class once.

jodosha:
@Bounga Merging this, thank you! :+1:

PR #48, opened 2015-06-25
Fixes the port on the link to demo app

LimeBlast:


jodosha:
@LimeBlast Thank you :+1:

PR #50, opened 2015-06-25
Update create-table.md

jejepage:
Typo and error fixes.

### DIFF
--- a/source/guides/migrations/create-table.md
+++ b/source/guides/migrations/create-table.md
@@ -6,18 +6,18 @@ title: Lotus - Guides - Migrations Create Tables
 
 ## Create Tables
 
-A table is defined via `#create_table`. This method accepts an two arguments: the **name** and a **block** that expresses the design.
+A table is defined via `#create_table`. This method accepts two arguments: the **name** and a **block** that expresses the design.
 
-Safe operation can be performed via `#create_table?`. It only creates the table if not exist.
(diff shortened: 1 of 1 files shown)


source/guides/migrations/create-table.md
```diff
@@ -6,18 +6,18 @@ title: Lotus - Guides - Migrations Create Tables
...
-Force operation can be performed via `#create_table!`. It drops the existing table and creates from scratch.
+Safe operation can be performed via `#create_table?`. It only creates the table if it doesn't exist.
+Force operation can be performed via `#create_table!`. It drops the existing table and creates a new one from scratch.
 **These operations shouldn't be used in migrations**.
 
 #### Column Definition
 
 To define a column we use `#column`, followed by the **name**, the **type** and **options**.
-The name must be an unique identifier within the table.
+The name must be a unique identifier within the table.
```
joneslee85: `u` is a vowel, thus it should be `an` not `a`

jejepage: Because the `u` is pronounced as a `y` (yoo nique), it's treated as a consonant. So it must be `a unique`. See http://wordwhirled.blogspot.be/2005/12/unique-or-unique.html or http://www.quora.com/Which-is-the-correct-grammar-usage-a-unique-or-an-unique.

PR #59, opened 2015-06-26
Fix validation specs

Bounga:
Validation specs weren't testing that the create action was really validating attributes but only that the user was redirected. A new spec ensure that the action is filled with expected errors.

It also fixes the spec which checks that the errors are displayed to the user. The current code can't work since `new` view doesn't know anything about `create` action. Both are isolated in tests. So I replaced it by a feature spec which can go through the whole process and really test for error displaying.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -832,6 +832,15 @@ describe Web::Controllers::Books::Create do
       response = action.call(params)
       response[0].must_equal 200
     end
+
+    it 'sets errors attribute accordingly' do
+      action.call(params)
+
+      refute action.params.valid?
+
+      action.errors.for('book.title').wont_be_empty
+      action.errors.for('book.author').wont_be_empty
+    end
   end
 end
 ```
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -932,6 +932,32 @@ describe Web::Views::Books::New do
 
     rendered.must_include('There was a problem with your submission')
     rendered.must_include('title is required')
+    rendered.must_include('author is required')
+  end
+end
+```
+
+We should also update our feature spec to reflect this new behavior:
+
+```ruby
+# spec/web/features/add_book_spec.rb
+require 'features_helper'
+
+describe 'Books' do
+  # Spec written sooner removed from brevity
+
+  it 'displays list of errors when params contains errors' do
+    visit '/books/new'
+
+    within 'form#book-form' do
+      click_button 'Create'
+    end
+
+    current_path.must_equal('/books')
+
+    assert page.has_content?('There was a problem with your submission')
+    assert page.has_content?('title is required')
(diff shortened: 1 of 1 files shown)


jodosha:
@Bounga It's not clear what's the intent here. Can you please explain a bit?

Bounga:
@jodosha My point is that in the original tests we duplicate params validation:

```ruby
class NewBookParams < Lotus::Action::Params
  param :book do
    param :title, presence: true
    param :author, presence: true
  end
end
```

So in fact we're not testing that our `apps/web/controllers/books/create.rb` controller is really doing its job and validating params. We're only testing errors rendering in isolation.

If you comment out the params validation in the controller, `spec/web/views/book/new_spec.rb` tests will still pass. In real world usage of the app, it won't have the expected behavior on missing params.

That's why I added `it 'sets errors in object'` test in `spec/web/controllers/books/create_spec.rb` which ensure that the controller does its job of validating params.

At this point we know that:

- `create` controller does params validation
- `new` view displays error messages if there are errors in params

My mistake in this PR is that I've deleted the `spec/web/views/books/new_spec.rb` explanations but should have kept it since it is needed to test that the view displays errors if any.

I also added the counterpart of the existing test in `spec/web/features/add_book_spec.rb`. I mean filling badly the form to see if the `new` template is rendered with error messages. I think that when you test something you should also test the case were it doesn't work especially if you handle it with custom behaviors (in our case, displaying errors).

Is it clearer? Does it makes sense to you to test that a controller is doing its params validation job?

If it does, I can bring back the `spec/web/views/books/new_spec.rb` part and remove the extra test in `spec/web/features/add_book_spec.rb` if you think it's too much.

To sum up, existing tests doesn't ensure that params validation are really handled at controller level and I feel like it's needed.

joneslee85:
@Bounga thanks for this PR. Can you please rebase this PR?

Bounga:
Yes, no problem. I think I need to bring back the spec/web/views/books/new_spec.rb first then rebase.

I'll do this as soon as possible

Bounga:
@joneslee85 Here it is

joneslee85:
@Bounga I am wondering that if we should test about error message in controller unit test? It is duplicated in our feature test anyway.

Bounga:
@joneslee85 It's its job to validate params so if I had to keep one test only I would keep the controller one. As I told before I "duplicated" the test in feature because the counterpart was here so I think if we cover one case (valid params) we should cover both

Bounga:
Final decision is up to you. You know better than me how a Lotus app has to be written and tested. Just tell me what you prefer and I'll do that.

joneslee85:
My preference is always outside-In, so I much prefer feature test to unit test. For this Getting Started guide, I think it is okay to have both just to demonstrate that unit testing in controller/view level is possible and easy to do, but also having feature test to test the end-to-end exp

PR #68, opened 2015-07-01
Change to proper redirect

tak1n:
In the example first `redirect_to '/books'` is used.

Then after knowing the routes helper it is replaced with `redirect_to routes.books_url`.

This should be `redirect_to routes.books_path`

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -981,7 +981,7 @@ We can use the `routes` helper method that is available in our views and actions
 We can make a similar change in `apps/controllers/books/create.rb`:
 
 ```ruby
-redirect_to routes.books_url
+redirect_to routes.books_path
 ```
 
 ## Wrapping Up


joneslee85:
according to the spec, it should be the full absolute path, that is `_url`

tak1n:
In https://github.com/lotus/lotus.github.io/blob/build/source/guides/getting-started.md#implementing-create-action a path is used for redirect and testing the redirect.

So it should be a `_path`

Also see this tests:

With `_url`

```
Run options: --seed 48976

# Running:

..F

Finished in 0.018131s, 165.4660 runs/s, 220.6214 assertions/s.

  1) Failure:
Web::Controllers::Books::Create::with valid params#test_0002_redirects the user to the books listing [spec/web/controllers/books/create_spec.rb:23]:
Expected: "/books"
  Actual: "http://0.0.0.0:2300/books"

3 runs, 4 assertions, 1 failures, 0 errors, 0 skips
```

With `_path`

```
Run options: --seed 2100

# Running:

...

Finished in 0.022746s, 131.8934 runs/s, 175.8579 assertions/s.

3 runs, 4 assertions, 0 failures, 0 errors, 0 skips
```

PR #79, opened 2015-07-27
Fix the failing test output in the Getting Started guide

andrii:
Welcome page is shown in all environments when no routes are defined.


andrii:
Closing since lotus/lotus#305 was merged.

PR #84, opened 2015-08-28
Use params from create action in spec

kholbekj:
It seem like the duplication is unnecessary. We are not supposed to be testing that the validations are the same as when the specs were written.


kholbekj:
Maybe it'd even make sense to cut the intermediate variable assignment?

joneslee85:
Thanks for this PR, normally yes, the params would be taken from the controller action, however in this case, we would like to avoid coupling with Controller as we focus on unit-testing this

PR #86, opened 2015-09-10
Various grammatical fixes and minor clarifications.

tinynumbers:
Fixed some unnecessary commas and minor grammatical issues.  Expanded a few things to further clarify.

### DIFF
--- a/source/guides/actions/control-flow.md
+++ b/source/guides/actions/control-flow.md
@@ -7,10 +7,10 @@ title: Lotus - Guides - Action Control Flow
 ## Callbacks
 
 If we want to execute some logic before and/or after `#call` is executed, we can use a callback.
-It's useful to declutter code for common tasks like checking if a user is signed in, set a record, handle 404 responses or tidy up the response.
+Callbacks are useful to declutter code for common tasks like checking if a user is signed in, set a record, handle 404 responses or tidy up the response.
 
 The correspoding DSL methods are `before` and `after`.
-They accept a symbol that is the name of the method that we want to call, or an anonymous proc.
+These methods each accept a symbol that is the name of the method that we want to call, or an anonymous proc.
 
 ### Methods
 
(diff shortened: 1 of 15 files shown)


source/guides/actions/exception-handling.md
```diff
@@ -81,12 +81,11 @@ module Web::Controllers::Dashboard
 end
 ```
 
-If we specify a symbol as the value for `handle_exception`, it will be used to manage the exception.
-In our case we want to protect the action from unwanted access: only admins are allowed.
+If we specify the name of a method (as a symbol) as the value for `handle_exception`, this method will be used to resond to the exception.
```
joneslee85: s/resond/respond

source/guides/actions/http-caching.md
```diff
@@ -56,16 +56,16 @@ end
 ## Conditional GET
 
 _Conditional GET_ is a two step workflow to inform browsers that a resource hasn't changed since the last visit.
-At the end of the first request, it receives special HTTP response headers to send back the next time it will come back.
-If that header matches the value that the server calculates, then the resource is still cached and a `304` status (Not Modified) is returned.
+At the end of the first request, the respons includes special HTTP response headers that the browser will use next time it comes back.
```
joneslee85: s/respons/responses

source/guides/actions/control-flow.md
```diff
@@ -153,7 +153,7 @@ module Web::Controllers::Dashboard
 end
 ```
 
-`#halt` accepts a HTTP status code as the first argument.
+`#halt` accepts an HTTP status code as the first argument.
```
joneslee85: s/an/a

tinynumbers: I think that could go either way - I (in American English) pronounce HTML as "aitch-tee-em-el" which indicates that "an" is proper.  See http://english.stackexchange.com/questions/25742/is-it-supposed-to-be-a-html-or-an-html for discussion.

PR #97, opened 2015-10-25
Fix minor typos

vdmgolub:


jodosha:
@vdmgolub Wow that's embarrassing :sweat_smile: Thanks for fixing it :+1:

PR #98, opened 2015-10-28
Fix documentation for nested resource routes

mattr:
Nested resource for single-to-plural and single-to-single showed routes for plural-to-plural and plural-to-singular. Replaced `/accounts/:account_id/` with `/account/` in each case so that the documentation shows the correct routes in each table.

Removed `:create_` from POST routes for nested resources (e.g. `:create_account_avatar` should just be `:account_avatar` with a POST method).

### DIFF
--- a/source/guides/routing/restful-resources.md
+++ b/source/guides/routing/restful-resources.md
@@ -263,7 +263,7 @@ end
     <td>/books/:book_id/reviews</td>
     <td>Books::Reviews::Create</td>
     <td>:create</td>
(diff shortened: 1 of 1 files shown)


joneslee85:
cc @AlfonsoUceda

jodosha:
@AlfonsoUceda PING :wink:

AlfonsoUceda:
@mattr sorry for the late response but it's ok!  merging!

PR #99, opened 2015-10-31
Add adapter if not using postgres

theunorthodoxdeveloper:
If the user decides to use a different adapter to `postgres` they will be thrown an error. This is also the case if the user sets up a new project without specifying an adapter.

In order to stop this happening, the adapter of their choice needs to be specified in the Gemfile, I have simply added a line to the get started documentation to outline this if the user decides to choose a different adapter they will need to add this to the Gemfile.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -398,6 +398,8 @@ The placeholders **_user_** and **_password_** should be replaced with the corre
 
 You may want to change `localhost` for `127.0.0.1` if you're using Linux.
 
+You may decide to use a different adapter, if you do make sure that you update your Gemfile to include the adapter of your choice and run bundle install.
+
 The database configured by default, called `bookshelf_development` running on `localhost`, should work fine for now.
 Lotus can create the database for us:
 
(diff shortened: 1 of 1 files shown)


joneslee85:
thanks for the PR, what do you think about listing all available adapter in a table?

theunorthodoxdeveloper:
Sounds good! I am unsure this should be on the get started page. I feel like it should be within the models documentation, we could then direct people to this information. What are your thoughts?

joneslee85:
sure, the started guide page should link to that page then

source/guides/getting-started.md
```diff
@@ -398,6 +398,8 @@ The placeholders **_user_** and **_password_** should be replaced with the corre
 
 You may want to change `localhost` for `127.0.0.1` if you're using Linux.
 
+You may decide to use a different adapter, if you do make sure that you update your Gemfile to include the adapter of your choice and run bundle install.
```
Nerian: May I suggest a different wording?

"You may decide to use a different adapter. If you do, make sure you update ..."

joneslee85: I am good with current wording

PR #100, opened 2015-11-05
fix mailer test

caspg:
Lotus::Mailer.deliveries returns an array. #last method is needed to
access mail object.


jodosha:
@caspg Thank you for this fix :+1:

PR #106, opened 2015-11-30
Remove words that presume knowledge

cllns:
"*Clearly*, *obviously*, *naturally*, and *of course* all presume too much about the reader's knowledge and perspective and often amount to excess verbiage." - [Wikipedia Manual of Style](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Words_to_watch#Editorializing)

Obviously (:wink:) we're not bound to Wikipedia's Manual of Style, but I think it's a good rule.

I left one 'actually', in the [Displaying Dynamic Data](https://github.com/lotus/lotus.github.io/blob/build/source/guides/getting-started.md#displaying-dynamic-data) of Getting Started, where it says *'If we run our feature test now, we'll see it fails — because our controller action does not __actually__ expose the books to our view'* because it seems useful.

### DIFF
--- a/source/guides/architectures/container.md
+++ b/source/guides/architectures/container.md
@@ -96,7 +96,7 @@ Web assets such as javascripts and stylesheets will be automatically served by t
 Our default application `Web` can be used as a UI interface for our customers.
 At a certain point in our story, we want to manage our users with an admin pane.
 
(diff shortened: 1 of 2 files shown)


jodosha:
@cllns Thank you. _Obviously_, as a non-native speaker, I'm really happy of fix my broken English. :smile:

PR #108, opened 2015-12-07
Add raw to JSON render

oana-sipos:
This is one of the only examples of JSON output using lotus. The latest versions of Lotus do HTML safety by default. By adding the `raw` to the method call we give people exploring the docs a more "ready to copy-paste" example.


joneslee85:
thx

PR #110, opened 2015-12-09
Change masculine pronouns to neutral

cllns:
We shouldn't use masculine pronouns by default. Instead we should be more inclusive by using neutral pronouns.


joneslee85:
thx

jodosha:
@cllns Thank you :+1:

PR #111, opened 2015-12-15
Remove wrong line

AlfonsoUceda:
Removed `http_caching true` line


jodosha:
@AlfonsoUceda Thank you! :+1:

PR #128, opened 2016-01-29
Fix syntax error in code snipper for custom helpers

katafrakt:
I spotted a minor syntax error in code snippet on page about custom helpers. This is a small fix for that.


katafrakt:
Found one more. Will commit in a sec.

jodosha:
@katafrakt Thank you very much! :+1: Have a great Friday!

PR #130, opened 2016-01-29
Explain how to use UUID as primary key instead of integer

bolshakov:


jodosha:
@bolshakov Thank you very much! :+1:

PR #131, opened 2016-01-29
Set grey color for command output in dev block

davydovanton:
### Before
screenshot 2016-01-29 20 12 14

### After
screenshot 2016-01-29 20 13 44

/cc @jodosha


jodosha:
@davydovanton Yeah, that looks better, thank you :+1:

davydovanton:
you're welcome :)

PR #135, opened 2016-01-30
Add information about generation action with --method

davydovanton:
Reference https://github.com/hanami/hanami/issues/489


davydovanton:
Thanks! 🎉

PR #143, opened 2016-02-01
Removes duplicate execution order statement

RossBarnie:
This line and the one after say exactly the same thing. I've removed this line because it is slightly less explicit than the alternative.


joneslee85:
thanks for the PR

PR #145, opened 2016-02-06
Add information about supporting link_to method

davydovanton:
Reference https://github.com/hanami/helpers/issues/48

/cc @joneslee85, @jodosha


joneslee85:
How about adding another example for 1 2 link_to(s) to demo that we have to concatenate them

davydovanton:
@joneslee85 yes, I like your idea :+1: 
I added more complicated example.

jodosha:
@davydovanton @joneslee85 Merged, thank you all!

davydovanton:
thanks you all :-1:

PR #146, opened 2016-02-06
Removed old generator's action parameter

hlegius:
In the past, it used to require the `--url=/path/to/action` parameter, but it doesn't modify anything anymore.


jodosha:
@hlegius @joneslee85 Thank you all :+1:

PR #152, opened 2016-02-13
improve pgsql syntax in example

wazoo:
I added 'IF NOT EXISTS' to the 'CREATE EXTENSION' line as that migration will fail if the extension is already created unless 'IF NOT EXISTS' is there as well.  It fails with an error as such:

```
/path/to/gem/lib/sequel/adapters/postgres.rb:186:in `async_exec': PG::DuplicateObject: ERROR:  extension "uuid-ossp" already exists (Sequel::DatabaseError)
```


jodosha:
@wazoo Thank you, merging this :+1:

PR #153, opened 2016-02-20
Fix typo in 'getting started' guide

iamrahulroy:
I was going through [getting started](http://hanamirb.org/guides/getting-started/) guide and I noticed a small typo. "It" should be "If".


jodosha:
@iamrahulroy Thank you very much! :+1:

iamrahulroy:
@jodosha Glad I could help.

PR #156, opened 2016-02-28
Update form helper documentation

davydovanton:
Add missing helper tags and inputs. Add documentation link to each heller. Add "Pass A Value" and "Share Code Between Views" examples.

### DIFF
--- a/source/guides/helpers/forms.md
+++ b/source/guides/helpers/forms.md
@@ -92,22 +92,24 @@ end
 
 ### Supported methods
 
-* color_field
-* date_field
-* datetime_field
-* datetime_local_field
-* email_field
-* hidden_field
-* file_field
-* fields_for
-* form_for
-* label
-* text_area
-* text_field
-* password_field
-* radio_button
-* select
-* submit
+* [check_box](http://www.rubydoc.info/gems/hanami-helpers/Hanami/Helpers/FormHelper/FormBuilder#check_box-instance_method)
+* [color_field](http://www.rubydoc.info/gems/hanami-helpers/Hanami/Helpers/FormHelper/FormBuilder#color_field-instance_method)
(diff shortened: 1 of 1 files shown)


source/guides/helpers/forms.md
```diff
@@ -201,6 +203,140 @@ Developers can customize attack handling.
 </form>
 ```
 
+### Pass A Value
```
jodosha: @davydovanton All this description was kept out with intention. This is overwhelming at the first glance and it's subject to change.

source/guides/helpers/forms.md
```diff
@@ -92,22 +92,24 @@ end
...
-* fields_for
-* form_for
-* label
-* text_area
-* text_field
-* password_field
-* radio_button
-* select
-* submit
+* [check_box](http://www.rubydoc.info/gems/hanami-helpers/Hanami/Helpers/FormHelper/FormBuilder#check_box-instance_method)
```
jodosha: @davydovanton This is good :+1:

source/guides/helpers/forms.md
davydovanton: yes, no problem. I added this part because I am often asked about share form code between views. What you think can we do something to help with this issue?

jodosha: @davydovanton I think that there is still room for improvement in views. By looking at this updated guide one may think "OMG, should I use all this stuff to create a simple form?". Again, this looks overwhelming. The right thing to do is to try to improve the API, instead of offer this informations overload.

davydovanton:
@jodosha deleted unnecessary examples

jodosha:
@davydovanton Thank you. Merging :+1:

davydovanton:
:tada:

PR #157, opened 2016-02-29
Getting Started: change 'either' to 'whether' in lead note

cllns:
I think this is what we want.

This also changes 'totally beginner' to 'total beginner', since I think that's more natural.


jodosha:
@cllns Thanks for fixing my :gb: :smile: :+1:

PR #159, opened 2016-03-01
Fix note about ZSH

cllns:
I think this is what we want.

I use `zsh` (5.0.6) and I don't have this problem. Is this note necessary? Has anyone else run into this? This note was added in #94.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -286,7 +286,7 @@ It gives us an empty action, view and template; it also adds a default route to 
 get '/books', to: 'books#index'
 ```
 
-If you're using ZSH, you may get `zsh: no matches found: books#new`. In that case, you can use:
+If you're using ZSH, you may get `zsh: no matches found: books#index`. In that case, you can use:
 ```
 % hanami generate action web books/index
 ```


joneslee85:
@cllns it could be on old versions, would you be able to investigate for us?

cllns:
Sure thing @joneslee85! I just looked into it and it seems like it's related to ['pattern removal'](http://zsh.sourceforge.net/Guide/zshguide05.html), but I can't figure out why I'm not having the problem.

cllns:
Hey @jasoncharnes! Could you help us out here? Can you try running 
```
hanami generate action web books#index
``` 
in a Hanami app? 
If you don't have `hanami` installed, you can try it with an old `lotus` app, it shouldn't matter here. (Lotus was renamed to Hanami recently).

Do you still get that same problem?
Can you run ` zsh --version`?

For completeness sake, mine says `zsh 5.0.6 (x86_64-apple-darwin14.0.0)`

jasoncharnes:
@cllns sure thing!

This seems to no longer be an issue with ZSH. (Hooray!)

I tested using ZSH `zsh 5.0.8 (x86_64-apple-darwin15.0)`

Let me know if you need anything else.

jodosha:
Folks, I use ZSH since before starting this framework and **never** ran into this problem.

I'm glad that this problem was solved for you as well. I'd leave this extra help because people can still experience this issue.

jodosha:
@cllns Thanks for this fix, merging it! :+1:

PR #161, opened 2016-03-08
Updated Nested Access section with a new Ruby 2.3 dig method examaple

paxer:


jodosha:
@paxer thanks for this PR. `Params` acts like `Hash`, because you're able to do `params[:id]`, but they aren't `Hash` instances, so you can't use `#dig` over there.

```ruby
(byebug) params.dig(:id)
*** NoMethodError Exception: undefined method `dig' for #<ParamsAction::Params:0x007fca723ab058>

nil
```

PR #165, opened 2016-04-01
Move shared head tags in partial and default zoom for tablets and phones

KELiON:
Hello!

In this PR I made two things:

### Default zoom
I was reading this docs yesterday on iPad and I always had to zoom in. In this PR I've added default zoom, that make displaying better on phones and tablets:

How it was on iPhone:

Now on iPhone:

How it was on iPad:

Now on iPad:
(no more empty white areas on sides)

### Move shared head tags to partial
2) I wanted to add one meta tag in `<head>`, but there were few layouts with almost equal tags inside `<head>`, so I moved everything in one partial and added this meta tag there. 

Feel free to merge it, if you like it. Btw, let me know if I can help with any small frontend stuff for hanami!


jodosha:
@KELiON Thank you for this PR. Just merged it :+1:

PR #169, opened 2016-04-25
Add warning about naming Controllers after Apps

fidothe:
Adds a warning to help people avoid / debug / fix constant lookup errors caused by using the same module name for an App and a Controller, as mentioned in hanami/hanami#561.

I'm a little unsure about multiple `<p class="warning">`s in a row, but in the GH preview output, wrapping the paras in a `<div class="warning">` just made the output vanish. (AFAICT there's nothing in the CSS that would cause a problem there in actual HTML, but I'm wary of introducing usage that just breaks useful reviewing tools.)


jodosha:
@fidothe Thanks, I'm merging this!

PR #175, opened 2016-05-12
Add information about custom CLI commands

davydovanton:
Reference to https://github.com/hanami/hanami/pull/566

### DIFF
--- a/source/guides/command-line/custom-commands.md
+++ b/source/guides/command-line/custom-commands.md
@@ -1,0 +1,31 @@
+---
+title: "Guides - Command Line: Custom Commands"
+---
+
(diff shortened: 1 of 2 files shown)


jodosha:
@davydovanton Hey thanks for this PR, but I think it should be expanded with real world example. This is a bare minimum "Hello World". The question we should always ask while writing guides is: how a developer that doesn't know nothing about this topic succeed in what they need to do?

Now let's put in their shoes, what if they want to generate some code like you did? Can you please expand with examples from your code? Thank you very much! ✨

davydovanton:
I think we can close this PR 👍

PR #177, opened 2016-05-16
Blog post: Validations Predicates

jodosha:


source/blog/2016-05-16-validations-predicates.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+        A name must be filled  and be a string and its size fall between 3 and 64.
+           👇            👇     👇        👇    👇       👇                👇    👇
+required(:name)      { filled?  &       str?   &      size?              (3 .. 64) }
+
+```
+
+Now, I hope you’ll never format code like that, but in this case, that formatting serves well our purpose to show how Ruby’s  simplicity helps to define complex rules with no effort.
+
+From a high level perspective, we can tell that input data for `name` is _valid_ only if **all** the requirements are satisfied. That’s because we used `&`.
+
```
solnic: Since many people will see this for the first time, maybe it'd be a good idea to mention macro system too for collapsing common predicate compositions? so explain that `required(:name) { filled? & str? & size?(3..64) }` can be collapsed to `required(:name).filled(:str?, size?: 3..64)`?

source/blog/2016-05-16-validations-predicates.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+When we check data, we expect only two outcomes: an input can be valid or not.
+No grey areas, nor fuzzy results.
+It’s white or black, 1 or 0, `true` or `false` and _boolean logic_ is the perfect tool to express these two states.
+Indeed, a Ruby _boolean expression_ can only return `true` or `false`.
+
+To better recognise the pattern, let’s get back to the example above.
+This time we will map the natural language rules with programming language rules.
+
+```
+        A name must be filled  and be a string and its size fall between 3 and 64.
```
solnic: you could add `Input must have a key called :name and its value must be...`?

source/blog/2016-05-16-validations-predicates.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+---
+
+For long time [Hanami::Validations](https://github.com/hanami/validations) had problems that we struggled to solve and new features were problematic to add.
+Data management is complex task with thousands of cases to cover and because validations deal untrusted input, edge cases are common.
+Even simple cases like _blank values_ management became an issue.
+
+We tried to fix these problems, but over the time we realized that we hit the limit of that syntax, which led to lack of flexibility for us and for developers themselves.
+
+At the same time [dry-rb](http://dry-rb.org) folks released a new, stronger validations gem: `dry-validation`.
+It chances, for the good, the way we express validation rules.
```
hieuk09: Is it a typo here: `changes` instead of `chances`?

source/blog/2016-05-16-validations-predicates.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+So we took the decision to radically change our syntax and to adopt `dry-validation` as a validations backend for us.
+
+## How It Will Work?
+
+`Hanami::Validations` will work with input hashes and lets us to define a set of validation rules **for each** key/value pair.
```
solnic: `lets` => `let`
`to define` => `define`

source/blog/2016-05-16-validations-predicates.html.markdown
jodosha: @solnic I wanted to present the rule as a high level business requirement instead of low level details (input, keys etc..).

solnic: Ah gotcha, makes sense

PR #183, opened 2016-05-24
refactor(guides) use data instead of hard coding

jooohn:
### description
Refactor guides toc to use data instead of hard coding all links.

### motivation
While reading a guide, I want to go to "next" (or "prev") page from the bottom of page like this (https://www.playframework.com/documentation/2.4.x/NewApplication), but they are hard coded. To generate the next page link automatically, I started with refactoring them. What do you think about this?

Thanks.

### DIFF
--- a/data/guides.yml
+++ b/data/guides.yml
@@ -1,0 +1,103 @@
+categories:
+  - path: /
+    title: Introduction
+    pages:
+      - path: getting-started
+  - path: architectures
+    pages:
+      - path: overview
+      - path: container
+      - path: application
+  - path: applications
+    pages:
+      - path: initializers
+      - path: rake
+  - path: routing
+    pages:
+      - path: overview
+      - path: basic-usage
+      - path: restful-resources
+        title: RESTful Resource(s)
+  - path: actions
+    pages:
+      - path: overview
+      - path: basic-usage
+      - path: parameters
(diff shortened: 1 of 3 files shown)


cllns:
This is really cool :)

jooohn:
Ah, sorry. I missed these methods (https://github.com/hanami/hanami.github.io/blob/build/config.rb#L109-125). However, they aren't used now, right (maybe because it )?

jodosha:
@jooohn Thanks for this PR. The idea is cool, but we shouldn't change the URL of the getting started guide.

jooohn:
@jodosha Thanks for review! I fixed this commit not to change URL.

data/guides.yml
```diff
@@ -0,0 +1,103 @@
...
+      - path: exception-handling
+      - path: control-flow
+      - path: http-caching
+        title: HTTP Caching
+      - path: share-code
+      - path: testing
+  - path: views
+    pages:
+      - path: overview
+      - path: basic
```
jodosha: This is `basic-usage`. 😄

jooohn:
Ah, sorry. I fixed it!

jodosha:
@jooohn Thanks, merging this! I'll wait for the PR that includes `previous` and `next` links ;)

PR #185, opened 2016-05-24
Add special page with status of all hanami parts

davydovanton:
I found this idea on [ROM site](http://rom-rb.org/status/). Also I'll be very glad to any comments and suggestions 💚 

### Desktop
screenshot 2016-05-25 01 59 56

### Mobile
screenshot 2016-05-25 02 00 04

/cc @hanami/core-team and @hanami/contributors

### DIFF
--- a/source/_footer.erb
+++ b/source/_footer.erb
@@ -9,6 +9,7 @@
           <li><a href="https://github.com/hanami" target="_blank">GitHub</a></li>
           <li><a href="https://twitter.com/hanamirb" target="_blank">Twitter</a></li>
           <li><a href="https://rubygems.org/gems/hanami" target="_blank">Rubygems</a></li>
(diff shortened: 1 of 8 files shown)


joneslee85:
I ❤️ this

pascalbetz:
nice.

jodosha:
@davydovanton I like the idea. 👍 

But we should link it elsewhere, not from the main navigation. There will probably be upcoming links over there, so better to keep it small for now.

davydovanton:
@jodosha what about footer?

davydovanton:
@jodosha I putted status link to footer, what do you think is it better now?

jodosha:
@davydovanton much better, thanks! 👍

davydovanton:
thanks for merge 🎉

PR #186, opened 2016-05-25
Same as #184, use <code> inside html tag

nanocity:
Sorry, I did not add whitespace to search and replace regexp so I missed some chunks.


jodosha:
@nanocity Thank you. Merging this.

PR #187, opened 2016-05-27
Update configuration guide for separately gem usage

davydovanton:
Thanks @flash-gordon for report.

/cc @hanami/core-team and @hanami/contributors


jodosha:
@davydovanton @flash-gordon This is implicit for all the Hanami gems. Guides are for full stack web projects, while single READMEs are for standalone usage.

flash-gordon:
@jodosha I believe I started with GH docs so things worked smoothly but then I accidentally switched to site's docs and was wondered why I get an exception. Thanks, you made things clear. @davydovanton told me the same, I just forget that I started with the GH page, too many open tabs in my browser, you know... :smile:

PR #188, opened 2016-05-27
Added Sao Paulo User Group into Community page

hlegius:
#### Motivation
I started a Local User Group here in São Paulo city (Brazil) to bring together all interested people in Hanami.

/cc @jodosha


jodosha:
@hlegius 💯 💚 💯

PR #147, opened 2016-02-09
Announcing Hanami v0.8.0

jodosha:
**This is a WIP to keep track of all the changes that we will introduce with v0.8.0**

  - [x] Revert https://github.com/hanami/hanami.github.io/pull/203

---

:bomb: **PLEASE DO NOT MERGE IT YET** :bomb:

PR #193, opened 2016-06-02
Update BookRepository methods calls

tacionery:
I update the BookRepositoy.method to BookRepository.new.method
According to the https://github.com/hanami/model/blob/master/README.md
The Repository methods are now instance methods instead of class methods

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -508,13 +508,13 @@ We can use Hanami's `console` command to launch IRb with our application pre-loa
 
 ```
 % hanami console
->> BookRepository.all
+>> BookRepository.new.all
 => []
 >> book = Book.new(title: 'TDD', author: 'Kent Beck')
 => #<Book:0x007f9af1d4b028 @title="TDD", @author="Kent Beck">
->> BookRepository.create(book)
+>> BookRepository.new.create(book)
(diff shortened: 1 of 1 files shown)


cllns:
Thanks! This should be merged when `hanami-model` 0.7.0 is released (possibly with `hanami` 0.8.0? #147)

source/guides/getting-started.md
```diff
@@ -538,8 +538,8 @@ describe 'List books' do
   before do
     BookRepository.clear
```
vasspilka: You forgot this one ^^

cllns:
Actually these changes were released with [`hanami-model` version `0.6.2`](https://github.com/hanami/model/blob/v0.6.2/lib/hanami/repository.rb) so this should be merged soon, since people are seeing invalid docs.

cllns:
@jodosha @joneslee85 can we get a merge on this as soon as possible? I'm afraid a lot of people are running into this issue. We've had 4 people come to the chat and ask about it in the past couple days. I'm sure many more ran into the issue as well (and didn't join the chat).

I think this change was supposed to be for `0.7.0`but it accidentally got released as `0.6.2`.

jodosha:
@cllns OMG, Thanks for spotting this problem. I'm gonna revert this merge and yank that release.

PR #196, opened 2016-06-03
class methods => instance methods

vasspilka:
Updates guides to be up to date with new model, changes where introduced in https://github.com/hanami/model/issues/290

### DIFF
--- a/source/guides/models/repositories.md
+++ b/source/guides/models/repositories.md
@@ -26,20 +26,23 @@ This architecture has several advantages:
 
 ## Interface
 
-When a class includes `Hanami::Repository`, it will receive the following interface:
+When a class includes `Hanami::Repository`, it will receive the following instance methods:
 
-  * `.persist(entity)` – Create or update an entity
-  * `.create(entity)`  – Create a record for the given entity
-  * `.update(entity)`  – Update the record corresponding to the given entity
-  * `.delete(entity)`  – Delete the record corresponding to the given entity
-  * `.fetch(raw)`  – Fetch raw datasets for the given raw query string (eg. SQL)
-  * `.execute(raw)`  – Execute raw command (eg. SQL)
-  * `.all`   - Fetch all the entities from the collection
-  * `.find`  - Fetch an entity from the collection by its ID
-  * `.first` - Fetch the first entity from the collection
-  * `.last`  - Fetch the last entity from the collection
-  * `.clear` - Delete all the records from the collection
-  * `.query` - Fabricates a query object
+  * `#persist(entity)` – Create or update an entity
+  * `#create(entity)`  – Create a record for the given entity
+  * `#update(entity)`  – Update the record corresponding to the given entity
+  * `#delete(entity)`  – Delete the record corresponding to the given entity
+  * `#fetch(raw)`  – Fetch raw datasets for the given raw query string (eg. SQL)
+  * `#execute(raw)`  – Execute raw command (eg. SQL)
+  * `#all`   - Fetch all the entities from the collection
+  * `#find`  - Fetch an entity from the collection by its ID
+  * `#first` - Fetch the first entity from the collection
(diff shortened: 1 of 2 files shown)


source/guides/models/repositories.md
```diff
@@ -26,20 +26,23 @@ This architecture has several advantages:
...
-  * `.delete(entity)`  – Delete the record corresponding to the given entity
-  * `.fetch(raw)`  – Fetch raw datasets for the given raw query string (eg. SQL)
-  * `.execute(raw)`  – Execute raw command (eg. SQL)
-  * `.all`   - Fetch all the entities from the collection
-  * `.find`  - Fetch an entity from the collection by its ID
-  * `.first` - Fetch the first entity from the collection
-  * `.last`  - Fetch the last entity from the collection
-  * `.clear` - Delete all the records from the collection
-  * `.query` - Fabricates a query object
+When a class includes `Hanami::Repository`, it will receive the following instance methods:
```
cllns: Can we change 'it will receive the following instance methods' to 'objects will receive the following methods'? (since it's the objects that get the methods, not the class)

vasspilka: @cllns Instance methods are methods of an object of a class, so the above is corrent.
I don't really like 'objects will receive the following methods' maybe put an 'its ' in the front...
Or maybe we could change it to "its objects will receive the following interface:" does that sound good?

cllns: oops yea, I meant to include the 'its"

so it should be: `When a class includes `Hanami::Repository`, its objects will receive the following interface:`

vasspilka:
Ok done

source/guides/models/repositories.md
```diff
@@ -26,20 +26,23 @@ This architecture has several advantages:
...
+  * `#fetch(raw)`  – Fetch raw datasets for the given raw query string (eg. SQL)
+  * `#execute(raw)`  – Execute raw command (eg. SQL)
+  * `#all`   - Fetch all the entities from the collection
+  * `#find`  - Fetch an entity from the collection by its ID
+  * `#first` - Fetch the first entity from the collection
+  * `#last`  - Fetch the last entity from the collection
+  * `#clear` - Delete all the records from the collection
+  * `#query` - Fabricates a query object
+
+Note that these methods are instance methods therefor are called after a new repository object has been created with
```
joneslee85: s/therefor/therefore

source/guides/models/repositories.md
```diff
@@ -96,6 +99,9 @@ This is a **huge improvement**, because:
 
   * If we change the storage, the callers aren't affected.
 
+Now we can get all books from an author ordered by date, and we can also suply an optional limit, with only one method call.
```
joneslee85: s/suply/supply

vasspilka:
spelling fixed

jodosha:
@vasspilka Thanks for your PR, but I have to revert it.

Our website MUST report the usage of the stable gems released, not the work that is happening in master. The reason is simple: a developer does: `gem install hanami` which gets them the stable version. Then they open the guides on the site and they look for instructions that concern code that wasn't released yet. This is misleading.

https://github.com/hanami/model/issues/290 will be part of the next release. This PR is correct (thank you again), but this wasn't the right time to merge it. The time will be on the next release day.

/cc @joneslee85 @cllns

vasspilka:
@jodosha Hello Luca, Sean @cllns said its already part of the realeases for a while now, since 6.2. 
Should I open the PR again?

jodosha:
@vasspilka I'm terribly sorry for the confusion, but that `0.6.2.1` release accidentally included that code. Until we will reach 1.0, we bump _minor version_ for breaking changes. Moving from class methods to instance methods **is** a breaking change, for this reason it will be included in 0.7.

This is the reason why I yanked that `0.6.2.1`.

/cc @joneslee85 @cllns

PR #198, opened 2016-06-05
that's --> that is

olimart:


cllns:
Thanks! I actually think `that's` is OK because the guides are written in an informal style.

But there are a couple grammar things that can be cleaned up here. Namely, `which` should come after a comma, but `that` shouldn't have a comma before it. http://writingexplained.org/which-vs-that-difference

How about: Browsers don't understand HTTP methods other than `GET` and `POST`. But Hanami embraces REST conventions, which goes beyond those two verbs. When we specify an HTTP method via the `:method` option, it adds a special hidden field named `_method`, which is understood by the application.

olimart:
looking good.

jodosha:
@olimart Thanks for this PR, but as @cllns said guides are written in informal style.
The other grammar suggestions are still welcome, but as this PR wasn't updated, I'm closing it.

Folks, feel free to open a new one with improved English. Thanks!

PR #208, opened 2016-07-25
Added Missing Instruction to v.0.8 upgrade

mudspot:
Added Instruction to v.0.8 upgrade that the variable adapter -> uri must be modified

Updated instruction to edit .env.* files for clarity.


cllns:
Thanks @mudspot! good catch :)

PR #210, opened 2016-07-31
Fix expected validation error messages

iBublik:
According to dry-v, 'is missing' message is given when there's no expected key and 'must be filled' is given when there's an empty value for that key.


iBublik:
Sorry, it's duplicate of #206

PR #214, opened 2016-08-10
Fix bold markup in Model Overview guide

roryokane:
Since the text “MUST” is within `<p>` HTML tags, it must be marked up with `<strong>…</strong>` HTML tags instead of `**…**` Markdown syntax.

That section on the live site: http://hanamirb.org/guides/models/overview/#custom-coercions


cllns:
Thanks!

PR #216, opened 2016-09-02
Fixing example in getting-started guide

guilherme:
I was just running each step of the guide to see how hanami works and found this little missing part on the bookshelf examples.

To make the test pass it  needs a div with an id equal to "books" wrapping each book div.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -607,6 +607,7 @@ Hanami is designed around simple objects with minimal interfaces that are easy t
 Let's rewrite our template to implement these requirements:
 
 ```erb
+<h2>All books</h2>
 <% if books.any? %>
   <div id="books">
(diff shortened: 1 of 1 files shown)


cllns:
good find. It looks like by "rewrite" it means just change that part of the template (rather than replace the whole thing). I'd be fine with putting the whole new template in there though (for copy-and-pasters 😄), or changing the text to say "edit" instead of "rewrite".

If we want to put the whole template, we should probably add the page headers too:
```
<h1>Bookshelf</h1>
<h2>All books</h2>
```

What do you think?

guilherme:
Good point. But the <h1>Bookshelf</h1> it's already  in the application layout. I've updated the file. see if it makes sense :)

cllns:
🎉 thanks!

Issue #215, opened 2016-08-19
Interactor Guide


cllns:
`Hanami::Interactor`, in [`hanami/utils`](https://github.com/hanami/utils) is a very useful component.

We should consider adding a Guide to the website about how to use it (and the benefits of it).

Perhaps it could be an extension of the `Getting Started` guide, working from the `bookshelf` example.

There's no `Hanami::Utils` section on the site, so not sure if that's worth adding.

Happy to help write/review this guide, if we choose to move forward with it.

vladdypwnz:
I'm on it, captn

joneslee85:
@cllns I agree, we should have a guide on that. Feel free to proceed.
@vladdypwnz please work with @cllns to ensure you two do not step on each other's foot. thanks for your offer 👍

vladdypwnz:
@joneslee85 Hey I have a pretty good draft I think. I'm new to helping out on github. How do I best submit it?

cllns:
Fork this repo, then add your draft, commit it (preferably to a branch for your sake, but `master` works too), and push it up. Then open a PR against this repo (and reference this issue, #215, in it)

cllns:
Closed via #299

PR #219, opened 2016-10-03
Added Utils Hanami::Interactor explanation with a real-world example. Indented some sloppily-formatted code.

vladdypwnz:
This is in reference to issue #215 .

This commit is useful because it adds documentation to the Guide about the Interactor, as there is no documentation on the Interactor at all besides the Rdoc (it's a little arcane and doesn't explain some critical methods). I've also cleaned up some sloppily pasted code.

The explanation is a bit long because the Interactor is a little bit tricky. 

However, I've cut about 50-75% from my original draft thanks to @cllns advice, let's see if we can make it even more palatable.

### DIFF
--- a/data/guides.yml
+++ b/data/guides.yml
@@ -100,6 +100,9 @@ categories:
       - path: assets
       - path: routes
       - path: version
+  - path: utilities
+    pages:
+      - path: interactor
   - path: upgrade-notes
     pages:
       - path: v060
--- a/source/guides/utilities/interactor.md
+++ b/source/guides/utilities/interactor.md
@@ -1,0 +1,125 @@
+#### What is an Interactor?
+
+`Interactor` is an extremely powerful module as part of `Hanami::Utils`. Let's look at some basics, a practical example and then see what makes the `Interactor` so useful and powerful.
+
+`Hanami::Interactor` is a take on a Service Object pattern. In short, in will encapsulate functionality for us.
+
+#### Overview
+`Hanami::Interactor` is a "wrapper" for functionality - it provides generic, automatic tools for interacting with it, hence the name.
+
+Because it's just a module, it can be `include`d in a class. It will give you access to the following methods and functionality:
+
+* `Interactor` class will call `self.valid?` after initialization unless `self.valid?` returns `true`. 
+> Note that `self.valid?` must return `true` or `false` - "truthy" or "falsy" returns will result in a failed status!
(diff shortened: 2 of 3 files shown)


cllns:
Thanks for working on this!

I'd like to see a simpler example, without `HTTParty`, `ReCaptcha`, or `Dry::Validation.Schema`. The example you gave is a good real-world example, but it's too complicated in my opinion. There's a lot going on. I think it's better to just focus on the `Interactor` itself.

The `SignUp` example here: http://www.rubydoc.info/gems/hanami-utils/Hanami/Interactor/Interface is nice & simple. (Would like to see one with `Book` instead though, for this guide)

We wouldn't want to encourage people to use `home#submit`. Instead, that should be `book#create`.

In general, I don't think we need to explain everything about Interactors. I'm thinking 3 sections: **Basic Information**, **Interface**, **Result**. The first would provide a basic example, and explain some benefits of this approach. The **Interface** section could explain `initialize`, `call` and `valid?`. The **Result** section could talk about `errors` and `success?`. 

And a very tiny thing, but we should use the `error!` style instead of `error!()`.

Let me know if you want me to take this over from here, or if you want to continue working on this!

vladdypwnz:
Feel free to take over if you have the time, it might be faster if you do it. I used an example of something I myself use. 

I disagree that we shouldn't explain everything about it - rdoc makes no mention for example, that `error()` will not stop execution. At least, not in the Interface example given or that `throw`s are suppressed inside `call`.

Since you have more experience with the tone of the Guide though, maybe it's best if you take over.

cllns:
Ok! 

If you've found errors or omissions in the RDoc, please submit a PR to `hanami/utils` to fix them :) The RDocs should be authoritative.

vladdypwnz:
Sounds good Sean.

This is probably a little off-topic but is there a Guide on using `Dry::Validation.Schema` at all in the Hanami Guide? There's the [new validation syntax](http://hanamirb.org/blog/2016/07/22/announcing-hanami-080.html#new-validations-syntax) kind of based on it. Would a guide on either be useful or did I already miss it in the guide?

cllns:
There's a guide for using `hanami/validations`, in the context of parameters: http://hanamirb.org/guides/actions/parameters/#usage

Going to close this PR. Thanks for working on this, will definitely be using this and our discussion as a basis for the PR I'll open. I'll make sure to get your input on it too :) 🌸

PR #221, opened 2016-10-03
gem install bundler, too!

cllns:
If any users are brand-new to the Ruby ecosystem, they probably won't have bundler installed. We want to be friendly to them too!


jodosha:
@cllns `bundler` is a [runtime dependency](https://github.com/hanami/hanami/blob/master/hanami.gemspec#L31) for `hanami`. On a fresh Ruby setup, by installing `hanami` you will get `bundler` too.

Why developers should specify both instead of a simpler `gem install hanami`?

cllns:
oops, I should've checked that first. 🙃

PR #238, opened 2016-12-02
Use SQLite in the Getting Started guide

### LINKED ISSUE #236: Getting Started - Remove Postgres in favor of SQLite
```
hanami new bookshelf --database=postgres
```
should just be
```
hanami new bookshelf
```

We don't use any advanced features of Postgres, and it makes the Getting Started guide more complicated. Some users may have to spend time installing Postgres (which can be a pain!) or dealing with authentication issues... all just to learn the basics of Hanami! 

I think we should include a note telling users that we're using SQLite and that in a production app they should use Postgres or some other production-ready database (and show them it's easy to use postgres instead: `hanami new bookshelf --database=postgres`)

TiteiKo:
Closes #236

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -68,6 +68,13 @@ Then we can use the new `hanami` executable to generate a new project:
 % hanami new bookshelf
 ```
 
+<p class="notice">
+  By default, the project will be setup to use a SQLite database. For real-world projects, you can specify your engine:
+  <code> 
+  % hanami new bookshelf --database=postgres
+  </code>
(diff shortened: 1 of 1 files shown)


cllns:
Thanks @TiteiKo! 
Can we also add an info bubble that says something like:

"For a production app, you'll probably want to run `hanami new bookshelf --database=postgres`" 

so that people know it's our recommendation?

TiteiKo:
@cllns Done

cllns:
Thanks! 🌸

PR #247, opened 2016-12-12
Fix Add a book spec title and comment

florish:
* Ensure spec label is consistent with previous example
* Clarify comment on omitting existing spec from code example


cllns:
Great, thank you @florish :)

PR #250, opened 2016-12-14
Add unit test for Book Repository

mzelenyuk:
Since we have a unit test for Book Entity it would be nice to add a unit test for Book Repository as well.


davydovanton:
thanks for contribution, but sorry, I think that this test is unnecessary, @cllns what do you think?

cllns:
I agree @davydovanton, this feels like testing the framework rather than testing the application.

Thanks for your PR though @mzelenyuk :)

PR #255, opened 2016-12-21
Clean up 'Getting Started' guide

cllns:
1. Standardize how we title our sections (we don't need to follow any particular style guide, but here's the APA's rule: http://blog.apastyle.org/apastyle/2012/03/title-case-and-sentence-case-capitalization-in-apa-style.html)

2. Update number of assertions / runs for tests, based on current Getting Started guide

3. Standardize `<!doctype HTML>` to `<!DOCTYPE HTML>`, since that's more popular and it's what our generators do.

4. Clarify what Gemfile does in simpler terms


jodosha:
@cllns 👍

PR #256, opened 2016-12-26
Add information how to test action flash messages

davydovanton:
/cc @cllns

### DIFF
--- a/source/guides/actions/testing.md
+++ b/source/guides/actions/testing.md
@@ -215,6 +215,31 @@ end
 Please be careful using doubles in unit tests. Always verify that the mocks are in a true representation of the corresponding production code.
 </p>
 
+
+### Flash messages
+
+In your action tests, you can check `flash` messages too. For this, you can use `exposures` method for getting all flash data.
+
+The following test example uses this method.
+
+```ruby
+# spec/web/controllers/users/create_spec.rb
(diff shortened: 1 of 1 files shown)


source/guides/actions/testing.md
```diff
@@ -215,6 +215,31 @@ end
...
 
+
+### Flash messages
+
+In your action tests, you can check `flash` messages too. For this, you can use `exposures` method for getting all flash data.
+
+The following test example uses this method.
+
+```ruby
+# spec/web/controllers/users/show_spec.rb
```
AlfonsoUceda: s/show_spec/create_spec

cllns:
This would've helped me last week :)

However, we don't mention `flash` anywhere else. It's experimental so we don't have it in the guides. Personally, I think it's time we add it to the guides with a warning. Don't think this should go in before that though. 💞 

Also, we should add a `sessions` testing guide, since it's extremely related: http://hanamirb.org/guides/actions/sessions/

AlfonsoUceda:
@jodosha I think we should remove experimental flag from flash, what do you think?

davydovanton:
agree with @AlfonsoUceda 👍

davydovanton:
merged, thanks for review @jodosha  👍

PR #258, opened 2016-12-29
Replace uses of `Web::Routes` with `Web.routes`

TrevorBramble:
It seems support for `Web::Routes` references was dropped (accidentally?) in 0.9.0 or later. This change updates the documentation to provide the style of reference that now works: `Web.routes`.

If this was an intentional change, we should probably also update the [upgrade notes](http://hanamirb.org/guides/upgrade-notes/v090/). I'm happy to include that in this PR if so.


jodosha:
@TrevorBramble Thank you very much! :+1:

PR #261, opened 2017-01-05
Add detailed information on mounting applications

### LINKED ISSUE #245: Be more specific about mounting applications
The [guides](http://hanamirb.org/guides/routing/overview/#mounting-applications) simply say:
> If we want to mount an application, we should use mount.

It would be helpful to specify that the `mount` instruction should go to `config/environment.rb`.

spaceghoul:
The mounting guide explains how to mount to a path and a subdomain but does not
explain where the mount code should live.  Rails mounts within the route config
but Hanami mounts within the environment config.  The difference is worth
documenting.

Fixes #245

### DIFF
--- a/source/guides/routing/overview.md
+++ b/source/guides/routing/overview.md
@@ -76,7 +76,16 @@ get '/rack-app', to: 'rack_app' # it will map to RackApp.new
 
 ### Mounting Applications
 
-If we want to mount an application, we should use `mount`.
+If we want to mount an application, we should use `mount` within the Hanami environment configuration file. The global configuration file is located at `config/environment.rb`. Place `mount` within the Hanami.configure block.
+
+```ruby
+Hanami.configure do
+  mount Web::Application, at: '/'
+  mount OtherApplication.new, at: '/other'
+
+  ...
+end
+```
 
 #### Mounting To A Path
 


cllns:
thanks! it seems a bit redundant with the `Mounting To A Path` section below it.

I wonder if we should re-organize this whole section. There are four important parts, as I see it:
- How to mount Hanami applications
- How to mount other Rack applications
- How to mount to a path
- How to mount to a subdomain

What do you think?

spaceghoul:
I think it is worth explaining 'where' to mount, then explaining 'how' and 'what' to mount in different ways.  With documentation, I would not focus on removing redundancy and err on the side of over-explaining.  I think you did point out two 'types' of mounting that are not discussed specifically and likely should be.  The 'what' you can mount isn't as directly talked about as the 'how' to mount.

I am just starting to get my feet wet with the documentation needs for Hanami and would like to see this specific issue addressed/closed and then I'll plan on expanding the mounting documentation as a follow-up.  Would that work for you?  It actually probably deserves its own detailed section outside of the overview.  Probably still within the routing section, but as an additional focused page within that section.

b264:
This is fantastic-- it is extremely common for me to find exactly "how" to do something from the documentation for whatever it is, but have no idea what file to put it in.  Including that is :100: :+1:

source/guides/routing/overview.md
```diff
@@ -76,7 +76,16 @@ get '/rack-app', to: 'rack_app' # it will map to RackApp.new
 
 ### Mounting Applications
 
-If we want to mount an application, we should use `mount`.
+If we want to mount an application, we should use `mount` within Hanami environment configuration file. The default configuration file is located at `config/environment.rb`. Place `mount` within the Hanami.configure block.
```
beauby: s/default/global/?

spaceghoul: I'm not sure what you are saying, it might be tied to something I'm unaware of but this is an attempt to solve your issue so I'm interested in your feedback.  Can you clarify?

b264: It's being suggested to replace string `default` with `global` here.  It's written in the style of a POSIX `sed` command.

spaceghoul: Hahaha I definitely looked at his comment as a path and not a string replace suggestion.  Hahaha.  Thank you and now I feel dumb.

cllns: @spaceghoul don't feel dumb! it's a silly arcane thing, and now you know about it :)

AlfonsoUceda: @spaceghoul could you amend what @beauby said? I'll be glad to merge it ;)

AlfonsoUceda: ping @spaceghoul

spaceghoul: Sorry everyone, was on a much needed long vacation.  I will update the PR.

spaceghoul:
@AlfonsoUceda this should be good now, updated the text and brought my fork up to date with changes while I was away.  Let me know if there is anything else.  Thanks.

AlfonsoUceda:
@spaceghoul thanks!

PR #263, opened 2017-01-13
Guides on Repositories: fix typo on source code example.

rmascarenhas:
That should be `repository`, if I'm not mistaken.


davydovanton:
Yep, you're right. Thanks for contribution 🌟

PR #264, opened 2017-01-19
Add gitter chat on index page

davydovanton:
Hey, I added [sidecar](https://sidecar.gitter.im) to our index page. I can add this to all pages but I think it will be unnecessary. What do you think?

### How it works
hanami-sidecar

/cc @jodosha @cllns @hanami/contributors

### DIFF
--- a/source/_footer.erb
+++ b/source/_footer.erb
@@ -55,3 +55,8 @@
   ga('create', 'UA-47369640-1', 'hanamirb.org');
   ga('send', 'pageview');
 </script>
+
+<script>
+  ((window.gitter = {}).chat = {}).options = { room: 'hanami/chat' };
+</script>
+<script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer></script>
--- a/source/stylesheets/application-minimal.css
+++ b/source/stylesheets/application-minimal.css
@@ -230,3 +230,17 @@ table.status th {
 table.status td {
   font-size: 16px;
 }
+
+.gitter-open-chat-button {
+  background-color: #534a7f;
+  border-color: #443d68;
+}
+
+.gitter-open-chat-button:focus, .gitter-open-chat-button:hover {
+  background-color: #48416f;
+  text-decoration: none;
+}
+
+.gitter-chat-embed {
+  z-index: 1000;
+}


cllns:
I like it! :)

AlfonsoUceda:
I like it but I don't know if it will be used, can we know this?

davydovanton:
@AlfonsoUceda I have no idea how we can know this. But we can revert this at any time.

AlfonsoUceda:
Yeah, just made the question because of curiosity ;)

davydovanton:
@AlfonsoUceda if we use google analytics we can create handler (or trigger) for this button I think

jodosha:
@davydovanton I'm intrigued. Yes we use GA. Do we want to set a goal to measure the success of this experiment?

pascalbetz:
I like the idea!

davydovanton:
@jodosha so, I think it will be the interesting experience. And after that, we can say work this solution or not. I believe that it will be helpful for other projects like rom, dry, TRB, etc.

Maybe it will be good to add this button to all pages. For example:
You read a guide about something. Find some complicated part, and you can ask help in one click instead of opening a new page, finding gitter chat, etc. WDYT?

jodosha:
@davydovanton I assumed that you wanted to add to all the pages.

davydovanton:
Also. I have the idea for another issue. I think we can set click link counter for other links like Twitter, Github, etc. It will be helpful for understanding how often users use this links and maybe we will drop some links after this experiment.

davydovanton:
As I said before it only on the index page. But I can add the button to other pages too. @jodosha do you think that we need to add this button to each page, yes?

jodosha:
> But I can add the button to other pages too.

Yes, let's try for a month.

jodosha:
> I think we can set click link counter for other links like Twitter, Github, etc. It will be helpful for understanding how often users use this links and maybe we will drop some links after this experiment.

Anything that helps to track how many people uses it. And possibly the percentage.

davydovanton:
@jodosha yep, I'll create the issue for GA improvements. Okay?

davydovanton:
Okay, I added sidecar button for all pages in our site 👍 
Let's try it 🎉

AlfonsoUceda:
LGTM 👍

davydovanton:
thanks for merge @AlfonsoUceda 👍

AlfonsoUceda:
you welcome ;)

Issue #269, opened 2017-01-24
'Download, Develop, Deploy in 5 minutes.' section doesn't work because heroku doesn't support sqlite


bbonislawski:
Some time ago `sqlite` was set to be default adapter. After that deploy on heroku doesn't work(since heroku doesn't support sqlite). Instruction should be updated to don't confuse newcomers which would see that it doesn't work

AlfonsoUceda:
@bbonislawski thanks for the issue ;)

AlfonsoUceda:
closed by #270

PR #270, opened 2017-01-24
set database to postgresql in download, develop, deploy in 5 minutes section

bbonislawski:
Setting db to postgresql fixes problems with deploy on heroku. Problem mentioned in #269

### DIFF
--- a/source/layouts/home.erb
+++ b/source/layouts/home.erb
@@ -102,7 +102,7 @@
 Successfully installed hanami-<%= hanami_version %>
 1 gem installed
 <br>
-<span class="command">% hanami new bookshelf</span>
+<span class="command">% hanami new bookshelf --database=postgresql</span>
 19 files successfully created
                       </code></pre>
                     </div>


oana-sipos:
Hey @bbonislawski, thanks for the suggestion. I can see your point, but instead of this approach, I would propose to mention that ”remember that sqlite does not work on Heroku, thus use postgresql (or another DB) for this task”. I can understand your point of frustration for beginners, but as frustrated as I was back then, it was still a very important lesson I had to learn. My 2 cents.

p.s. There is already a comment on the getting started guide which suggests to generate a project using a different DB for real projects: http://hanamirb.org/guides/getting-started/

bbonislawski:
I understand decision for sqlite in guide, but user sees `Download, Develop, Deploy in 5 minutes. ` on home page and then he will see, that it won't work even after copying commands from home page. It's not a problem for me but could be problem for newcomerse.

cllns:
@oana-sipos I disagree, I think this PR should be merged. We have the code there so people could copy & paste or enter as-in. We should provide the exact commands that will work. The comment on the Getting Started guide is good and necessary, but this is a different example. Thoughts?

b264:
I definitely think this should be merged

AlfonsoUceda:
@bbonislawski mysql works on Heroku? I think we could use the two version
```ruby

cllns:
@AlfonsoUceda I think we should keep it as simple as possible, and not give people the option.
It looks like it's not supported very well on Heroku. [Their guide](https://devcenter.heroku.com/articles/heroku-mysql) is about migrating to Postgres, though they do offer 3rd party addons for MySQL.

AlfonsoUceda:
@cllns then that is enough ;)

PR #271, opened 2017-01-26
Fix typo on views basic usage page

gizotti:


gizotti:
@davydovanton my pleasure. 👍

PR #279, opened 2017-02-18
Fix sample code in Action Parameters guide

nickgnd:


davydovanton:
nice catch, thanks! 👍

PR #281, opened 2017-02-23
Update documentation for repository last and second methods

davydovanton:
Reference to https://github.com/hanami/model/issues/380

/cc @hanami/documentation @cllns @jodosha

### DIFF
--- a/source/guides/models/repositories.md
+++ b/source/guides/models/repositories.md
@@ -242,42 +242,58 @@ class BookRepository < Hanami::Repository
 end
 ```
 
-## First and last record
-We are used to that `#first` and `#last` methods called for ORM object returns first and the second
-record in a query. But it's a not good practice. It's common because it exists in `Array` and popular
-ORMs provide it too, the problem is that you can't tell what the order is, and you can't possibly
-expect to get consistent results when an order is not enforced in any way.
+## Retrieving the first and last records
+We're all used to calling the `#first` and `#last` methods on `Array`.
+Our repositories provide these methods, but we recommend *not* using them.
 
-That's why if you want to use `#first` or `#last` methods in repository query you need to use
-`#one` and `#one!` methods instead:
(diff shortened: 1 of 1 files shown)


cllns:
Thanks for remembering to do this @davydovanton <3

I committed some changes to this branch. I think they help explain the rationale behind recommending `one` vs `first`. 

**But** I don't think this should be merged now:
- someone else should check my changes for accuracy, and
- we should add note about how `first` and `last` are different (returning `Hash` vs entity). I feel like that was the impetus behind this change, and we don't directly address it. Instead, we tell people how to avoid it.

As a bigger question, before `1.0.0.rc1`, should we:
1. remove `first` and `last` from `Hanami::Repository` altogether? Instead we could encourage `limit` & `one` (and, optionally, `order`). @solnic [mentioned considering removing](https://github.com/rom-rb/rom-sql/issues/64#issuecomment-275660022) `first` and `last` from `rom-sql`, and if we can be ahead of this change, it could be good :) If so, should combine with 3.
2. change `first` implementation in `Hanami::Repository` to `root.limit(1).one` (and `last` to `root.limit(1).reverse.one`), to avoid the Hash/entity issue https://github.com/hanami/model/issues/380#issuecomment-280253302
3. add a warning to `first` and `last`, discouraging their use, in favor or `limit` and `one`?

If it's too late for any of these, that's fine. Just thought I'd bring it up :)
@hanami/core @jodosha

davydovanton:
@cllns 
raw

AlfonsoUceda:
@cllns @davydovanton good work ;)

The documentation is fine for me but as @cllns said:
> we should add note about how first and last are different (returning Hash vs entity). I feel like that was the impetus behind this change, and we don't directly address it. Instead, we tell people how to avoid it.

This would be perfect :)

AlfonsoUceda:
@cllns @davydovanton is there a chance to finish this?

PR #282, opened 2017-02-27
Add instructions about how to setup test database

Bounga:
This commit is intended to help people with setting up their database to run their tests.

Hanami is all about TDD / BDD and there's a strong emphasis on how to write tests but there's a missing part about how to configure the test database so new comers can be stuck with an error message telling them that a database is missing even if they followed the getting started guide step by step.

With these explanation the gap is filled.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -159,9 +159,11 @@ Note that, although Hanami is ready for a Behavior Driven Development workflow o
 We'll go with [Minitest](https://github.com/seattlerb/minitest) here (which is the default), but we can use [RSpec](http://rspec.info) by creating the project with `--test=rspec` option.
 Hanami will then generate helpers and stub files for it.
 
-Nonetheless we'll have to setup the test database to be able to run the tests.
-First we have to define our database url in `.env.test`.
-Now our database is configured we have to create it and migrate our schema by running
+<p class="notice">
+  Please check .env.test in case you need to tweak the database URL.
+</p>
+
+We have to migrate our schema in the test database by running:
 
 ```shell
 % HANAMI_ENV=test bundle exec db prepare
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -70,7 +70,7 @@ Then we can use the new `hanami` executable to generate a new project:
 
 <p class="notice">
   By default, the project will be setup to use a SQLite database. For real-world projects, you can specify your engine:
-  <code> 
+  <code>
   % hanami new bookshelf --database=postgres
   </code>
 </p>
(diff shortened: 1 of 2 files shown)


source/guides/getting-started.md
```diff
@@ -159,6 +159,16 @@ Note that, although Hanami is ready for a Behavior Driven Development workflow o
 We'll go with [Minitest](https://github.com/seattlerb/minitest) here (which is the default), but we can use [RSpec](http://rspec.info) by creating the project with `--test=rspec` option.
 Hanami will then generate helpers and stub files for it.
 
+Nonetheless we'll have to setup the test database to be able to run the tests.
+First we have to define our database url in `.env.test`.
```
cllns: Will users have to define `DATABASE_URL` in `.env.test`? It should already be set for them

Bounga: Yup you're right, it already exists. What I wanted to say is that you'll maybe want to tweak it. The default is: `"postgresql://localhost/shopit_api_test"` but some people do set a password on it, even if it's a local development DB.

jodosha: @Bounga Can we change this into a _notice_ block? Something like "Please check `.env.test` in case you need to tweak the database URL". Thanks 😄

source/guides/getting-started.md
```diff
@@ -159,6 +159,16 @@ Note that, although Hanami is ready for a Behavior Driven Development workflow o
...
+Now our database is configured we have to create it and migrate our schema by running
+
+```shell
+% HANAMI_ENV=test bundle exec db prepare
```
jodosha: @Bounga `db prepare` is `db drop` + `db create` + `db migrate`. Because we already have created the database this should be `migrate` instead of `prepare` 😉

source/guides/getting-started.md
Bounga: Yes sure, nice reword.

source/guides/getting-started.md
Bounga: I can't find anything in the getting started guide that states we have to do `hanami create db`. In the guide it work because we make the assumption that the reader uses sqlite.

So I don't think we've already created the database. Do we have to be explicit about that and add something saying that if you're using a "real" DB, you have to create the DB?

And even if the user has created the DB, I think that only the development one has been created, he still has to create the test one or do I miss something?

jodosha: @Bounga My apologize. At a first look, I confused the changes you made for `source/guides/command-line/database.md` as included in the getting started guide. In the first file you added `db create`, and I assumed it was the same guide.

This is good to go. Thank you!

Bounga:
@jodosha Here's a new version with a more explicit explanation about the need to check `.env.test` file. If all is ok for you, do you want me to rebase my commit into one?

jodosha:
@Bounga Thank you I squashed and merged the commits by myself. Thanks for this contribution! 👍

Bounga:
@jodosha That's my pleasure, really. Love your work.

PR #292, opened 2017-03-09
Removed misleading sentence from opening paragraph

graywolf:
The sentence I removed gived me impression that I should learn rails first and then come back to hanami. From reddit I understand that this is not correct. I'm intersted in hanami even when I never touched rails.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -16,7 +16,6 @@ title: Guides - Getting Started
 
   <p>
     I warn you that whether you're a total beginner or an experienced developer <strong>this learning process can be hard</strong>.
-    After 10 years you develop a way of working, and it can be painful for you to change. However, <strong>without change, there is no challenge</strong>.
   </p>
 
   <p>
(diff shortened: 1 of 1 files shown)


davydovanton:
Sorry, but I don't see any negative in this sentence.
It's about changes your mind and not about rails. Also, I don't find any `rails` matches on this page, that's why I have a question: why do you think that this about rails and not Sinatra or something else?

cllns:
I see both of your points. I think this line is important, but it can be phrase better. How about something like:

> Over time, we build expectations about how things _should_ be, and it can be painful to change. But **without change, there is no challenge** and without challenge, there is no growth.

?

graywolf:
@davydovanton 
> why do you think that this about rails and not Sinatra or something else?

sinatra isn't 10 years old yet. since rails is about 11, I reached the impression that it was reference to rails. but true, I might be wrong, what is the "after **10** years" refering to then?

@cllns 
> I see both of your points. I think this line is important, but it can be phrased better.

I like it, I'll update the pull request.

AlfonsoUceda:
@graywolf thanks for this PR, folks I like it this PR what do you think @hanami/core ?

davydovanton:
much better now 👍

cllns:
👍

davydovanton:
@jodosha could you check this? Thanks!

davydovanton:
thanks for this contribution 👏

PR #298, opened 2017-03-25
Update Getting Started guide

onshi:
As I read Hanami's Getting Started guide I found it slightly outdated. This PR is attempt to make it up to date. I hope you are ok with this.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -448,6 +448,12 @@ describe Book do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


cllns:
Hi there! Thanks for this PR, but I think the issue is that you're using an older version of Hanami. Starting in 1.0.0.beta1, we generate a migration when running `hanami generate model`. Can you confirm that you're using the version or newer?

See: https://github.com/hanami/hanami/pull/682

Going to close this but please let me know if you still have the problem after upgrading, or you're already on the most recent version of `hanami`. 🌸

PR #308, opened 2017-04-06
Add map with all core developers to team page

davydovanton:
<details> 
  <summary> Full team page screenshot </summary>
  
</details>

### DIFF
--- a/source/stylesheets/application-minimal.css
+++ b/source/stylesheets/application-minimal.css
@@ -259,14 +259,11 @@ table.status td {
 
 .core-map {
   margin: 50px 0 10px;
-  width: 100vw;
+  width: 100%;
+  max-width: 1024px;
   position: relative;
-  left: 50%;
-  right: 50%;
-  margin-left: -50vw;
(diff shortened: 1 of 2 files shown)


cllns:
- Made the icons smaller and higher resolution (so they're not blurry)
- Rounded out the co-ordinates, since 2 decimal points is sufficient (especially for rough co-ordinates!)
- Made the map static and removed a lot of the controls. No zooming or panning (so the map can't "hi-jack" the mouse in any way)

<details> 
  <summary> Update full team page screenshot </summary>
  
</details>

Note: I am still getting a bug, I see there's supposed to be a mouseover that shows the name of the person, but I get a JS "d.type is undefined" error. Briefly tried to figure it out but I couldn't.

AlfonsoUceda:
@davydovanton sorry for the late response. I think we don't need this page, I know it is cool but it doesn't give value to the page.

davydovanton:
yep, I agree. Let's close PR 👍

PR #310, opened 2017-04-06
Fix doc about model generator

y-yagi:
The model generator also creates a migration.


cllns:
Good find, thanks @y-yagi 🌸

y-yagi:
Thanks! :cherry_blossom:

PR #313, opened 2017-04-11
Use a word 'initialized' instead of 'initialised'

ryu39:
Thank you for your nice documents.

I found a small improvement in the Getting Started page.

I think it is better to use a word `initialize` instead of `initialise` because the previous description (... which we pass to initialize params) uses it.
Also, `initialize` is a very popular method name in Ruby.


cllns:
Thanks! 🌸

ryu39:
Thank you for merging!!

sakura

PR #315, opened 2017-04-14
Fix some typos in the Mailer Basic Usage guide

paulcsmith:
There were a couple small typos in the mailer guide. I hope this helps 👍 

Congrats on 1.0!

### DIFF
--- a/source/guides/mailers/basic-usage.md
+++ b/source/guides/mailers/basic-usage.md
@@ -24,7 +24,7 @@ end
 ```
 
 <p class="warning">
-  Both <code>.from</code> and <code>.to</code> informations MUST be specified when we deliver an email.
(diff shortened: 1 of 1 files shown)


source/guides/mailers/basic-usage.md
```diff
@@ -24,7 +24,7 @@ end
 ```
 
 <p class="warning">
-  Both <code>.from</code> and <code>.to</code> informations MUST be specified when we deliver an email.
+  Both <code>.from</code> and <code>.to</code> MUST be specified when we deliver an email.
```
paulcsmith: I removed the word "informations" because it felt redundant. If you want me to add it back I can, but I'll add it as "information" because I believe that is the correct way to write this

cllns:
Woo hoo! thank you 🙌 🌸

PR #317, opened 2017-04-15
Mention the need to defined params in exposures in tests

radanskoric:
I'm working on my first project in Hanami and I ran into this same issue quite quickly: https://github.com/hanami/helpers/issues/71

Since it took me a bit of time to figure out what is going on I thought it might be avoided for people trying hanami in the future by tweaking the guides a bit.


davydovanton:
I Agree, it's common problem. Thank so for you work 👏

PR #320, opened 2017-04-25
avoid test failure by using books_path

tudorpavel:
By following the guide and using the `books_url` routes helper, instead of `books_path`, the action test fails:

```
  1) Failure:
Web::Controllers::Books::Create::with valid params#test_0002_redirects the user to the books listing [/home/tpavel/workspace/bookshelf/spec/web/controllers/books/create_spec.rb:27]:
Expected: "/books"
  Actual: "http://0.0.0.0:2300/books"
```

I see `books_path` is used in the corresponding commit in the [hanami/bookshelf repo](https://github.com/hanami/bookshelf/commit/7a99bb50a87449d2e9794a2498de751f4fc9b4a7), so this PR should keep things consistent.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -1073,7 +1073,7 @@ We can use the `routes` helper method that is available in our views and actions
 We can make a similar change in `apps/web/controllers/books/create.rb`:
 
 ```ruby
-redirect_to routes.books_url
+redirect_to routes.books_path
 ```
 
 ## Wrapping Up


cllns:
Thank you :) I've been meaning to update the `bookshelf` repo but this works better IMO.

URL's used to be preferred for redirects (based on an IETF RFC) but now paths are OK.

tudorpavel:
You're welcome. :)

Yes, I thought about changing the test to use a regex matcher and keep books_url, but it isn't worth complicating a Getting Started guide.

jodosha:
@tudorpavel Thanks for this PR, I think that if we use `books_path`, we should make this expectation stricter:

From:

```ruby
response[1]['Location'].must_include '/books'
```

To:

```ruby
response[1]['Location'].must_equal '/books'
```

tudorpavel:
@jodosha, the [second version](https://github.com/hanami/hanami.github.io/blob/build/source/guides/getting-started.md#securing-our-form-with-validations) of the test actually uses `must_equal` and it is the test that was failing for me after using the `books_url` helper.

Indeed, the [first version](https://github.com/hanami/hanami.github.io/blob/build/source/guides/getting-started.md#implementing-create-action) of the test is using `must_include` and we can use `must_equal` there as well. I'll do another quick PR.

cllns:
good find @jodosha :)

PR #321, opened 2017-04-25
Add hint that mailers folder is now required

kaikuchn:
See https://github.com/hanami/hanami/issues/767
@jodosha turns out one could also remove the line specifying the mailer root folder.

I think a simple inline hint should suffice. WDYT?


jodosha:
@kaikuchn Thank you very much! 👍

PR #322, opened 2017-04-26
use consistent expectation for redirect location

tudorpavel:
Related PR: #320

As discussed in previous PR, the expectation for redirect location should be strict and consistent across versions of the test before and after refactoring.


jodosha:
@tudorpavel Wow, thanks for the quick PR. Merging it! 👍

tudorpavel:
My pleasure. :slightly_smiling_face:

Issue #314, opened 2017-04-11
Remove bundle directory


radar:
The `bundle` directory is quite large and takes a long time to clone down here in Kangaland. I would suggest removing this directory and instead relying on the `Gemfile.lock`. 

My concern with it being so huge is that it is a barrier-to-entry for new OSS contributors, who may also come from countries with slow internet connections.

WDYT?

b264:
I also notice it has darwin extensions in it and not everyone is running their POSIX on darwin

tudorpavel:
@radar, my thoughts exactly. Surely the `bundle` directory doesn't need to be inside the repo.

From the #140 PR it seems to me it was a solution for the Travis [build failure](https://travis-ci.org/hanami/hanami.github.io/builds/105964885) when building the sitemap.

graywolf:
I don't understand the request. Even if the dir is deleted (which it was in #389 ), it's still part of repo history and therefore downloaded when cloning. Unless I'm missing something fundamental about git here.

We could rewrite histore and remove it complete but not sure it that's worth it.

radar:
If you removed it now then a shallow clone would be fast. As I said in OP, it would significantly lower the bar to entry.

radar:
With it deleted in the other issue, then this fixes my issue.

PR #323, opened 2017-04-26
remove bundle folder

tudorpavel:
Relates to: #314, #140

I activated Travis on my fork and the middleman build passed without the `bundle` folder. @joneslee85, was there any other reason to commit the `bundle` folder?


cllns:
Looks like we might have to upgrade to middleman 4, in order to pin the right version for the sitemap gem? https://github.com/Aupajo/middleman-search_engine_sitemap/issues/5#issuecomment-200246974

tudorpavel:
> Looks like we might have to upgrade to middleman 4

Yep, I saw that but I wasn't sure how to test everything after the upgrade. From the previous PR #140, I assumed the `bundle` folder was added only because of problems on Travis, so that's why I tried the build without it on Travis and the [sitemap was generated successfully](https://travis-ci.org/tudorpavel/hanami.github.io/builds/226071960#L529).

I think @joneslee85 is best equipped to confirm that the `bundle` folder was added only because of Travis.

cllns:
We use Travis to build the static-site :/

tudorpavel:
Yup, I know and I'm saying the sitemap build worked on Travis on my fork without the bundle folder, so I think it's not necessary to have the bundle folder in git anymore.

cllns:
@davydovanton Can you look into this? Would certainly make the build process simpler, to remove the local copies of gems in the `bundle/` folder

AlfonsoUceda:
Closing this PR because it is stale.

PR #324, opened 2017-04-27
Fix example for check constraint code

davydovanton:
I have migration:
```ruby
Hanami::Model.migration do
  change do
    create_table :coupons do
      primary_key :id

      column :type,   String
      # ..

      check %(type IN('fixed', 'percent'))
    end
  end
end
```

## With double quotes
```
[hanami] [INFO] (0.000143s) ROLLBACK
/Users/anton/.rvm/gems/ruby-2.3.0/gems/sequel-4.45.0/lib/sequel/adapters/postgres.rb:191:in `async_exec': PG::UndefinedObject: ERROR:  type "str" does not exist (Sequel::DatabaseError)
```

/cc @hanami/core

### DIFF
--- a/source/guides/migrations/create-table.md
+++ b/source/guides/migrations/create-table.md
@@ -145,6 +145,6 @@ create_table :users do
   column :role, String
 
   check { age > 18 }
-  check %(role IN("contributor", "manager", "owner"))
+  check %(role IN('contributor', 'manager', 'owner'))
 end
 ```


cllns:
Huh! weird. Should we document this somewhere? (That you need to use double quotes)

AlfonsoUceda:
@davydovanton that `check` method is from sequel o rom-sql?

davydovanton:
@cllns I think no. We add this case to documentation example and I hope it will be good 👍 
@AlfonsoUceda good question. It's a migration method that's why I think (I didn't check it) that the method from sequel

davydovanton:
@cllns @AlfonsoUceda WDYT can we merge this? :)

AlfonsoUceda:
@davydovanton yeah no problem

PR #325, opened 2017-04-28
Instruct to prepare test db before entity tests

kaikuchn:
As @sheldonh in the hanami chat mentioned, the instruction in the getting started guide do not explicitly state to prepare the test database. When one sticks to the guide the test therefore fails.

See [his message from 28. April 2017 14:14](https://gitter.im/hanami/chat?at=5903320b12d2409935a8c34d)

Since both the development and test database need to be prepared for the subsequent parts of the getting started guide to work properly, I suggest to apply the changes of this PR.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -409,10 +409,11 @@ Hanami provides a DSL to describe changes to our database schema. You can read m
 about how migrations work in the [migrations' guide](/guides/migrations/overview).
 
 In this case, we define a new table with columns for each of our entities' attributes.
-Let's prepare our database:
+Let's prepare our database for the development and test environments:
 
 ```
 % bundle exec hanami db prepare
+% HANAMI_ENV=test bundle exec hanami db prepare
 ```
 
 ### Working With Entities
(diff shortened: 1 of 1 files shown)


source/guides/getting-started.md
```diff
@@ -409,10 +409,11 @@ Hanami provides a DSL to describe changes to our database schema. You can read m
 about how migrations work in the [migrations' guide](/guides/migrations/overview).
 
 In this case, we define a new table with columns for each of our entities' attributes.
-Let's prepare our database:
+Let's prepare our database for the develop and test environments:
```
sheldonh: I think `develop` should be `development`.

cllns:
good idea! thanks for this. Can address @sheldonh's comment? Then we can merge 🌸

cllns:
Also, can you edit/remove lines 499-504? They do this same thing, further down in the guide, but people don't need to do it twice

kaikuchn:
How about now?

cllns:
Great, thank you @kaikuchn! 👍

PR #328, opened 2017-05-05
Fixed grammatical errors in projects/initializers.md and routing/overview.md

derekpovah:
"Or" makes the subject a singular subject, so the helping verb "is" should be used.


cllns:
Thank you! 🌸

PR #330, opened 2017-05-11
Full-text search

mbajur:
Hello!

I was missing one simple thing on hanami documentation website - ability to search the docs fast. I also wanted to investigate full-text searching features for middleman projects and decided to combine these two needs of mine and try to implement such feature for hanami website. This is how it looks:

I know that i was not sharing my plan on implementing this anywhere nor confronted my idea with anyone so it's totally fine if you don't want to use it. However, if you feel it can be usefull, i still have few things left to make it "production ready"

- **Improve the design?** I'm not a designer so probably someone with better design skills will want to improve the look of it
- **Bundle the assets** - i'm using CDN for external libraries like lunr.js. They probably should be bundled with the middleman app
- **Improve indexing** - right now, full html output is getting indexed in the search.json index - it would be better to sanitize it and remove the html tags
- **Not perfect** - what i've done is doing it's job pretty fine but it's not perfect. For instance - it can find any word in the guide page content but it can't redirect user directly to it. Improving this would involve changing the way of indexing the files completely. It's doable but problematic so i'm leaving it for later.

If you're interested in merging the request, i can take care of all the points above instead of the design one :)

Cheers!


mbajur:
Changed display of results a bit (removed `Guides - ` prefix and added `Guides` group heading)

image

cllns:
😍😍😍

We have plans to add a site for central API docs (so they're all together instead of on different pages). Not sure if this should integrate with that or how, but regardless, I like this a lot.

mbajur:
Is that site with API docs middleman-powered just like current hanami.github.io ? If so, that can be easily integrated :)

cllns:
@mbajur Looks like it will be! https://github.com/hanami/docs (url doesn't work because it's not complete yet)

davydovanton:
@mbajur great idea! Thanks for implementation.
About docs: unfortunately, it's not a middleman project. We use yard for generating it.

@cllns we have search field in docs :)
screenshot 2017-05-12 12 34 21

(you can check it here http://hanamirb.org/docs/1.0.0/)

mbajur:
Ok but are you guys planning to move the main site from middleman to yard as well?  Or will they exist in parallel.

I think it could be somehow possible to add a search source pointing to the yard docs as well if that's your concern.

Anyway - what's the plan then? You want to use this search thing I did (if so I'm gonna close the remaining tasks) or should I drop it? :)

cllns:
If the searches have to be separate, that's fine. I still think it's valuable having search for the Guides. What do others think? @hanami/core

oana-sipos:
👍  to adding search for the Guides.

AlfonsoUceda:
@mbajur thanks awesome! :) We'll have both sites, docs in yard for the APIs and guides web for tutorials.

I think for accepting this PR and merge it we need at least the redirect option. We can't merge the PR without this feature because it would be uncompleted. What do you think? can you provide that? Thank you so much.

mbajur:
@AlfonsoUceda what you mean by redirect option?

AlfonsoUceda:
@mbajur I mean, click in the box and be redirected to that page, is that possible?

mbajur:
Ohh this one! It's already pushed since the beginning. I just didn't shown it in the recorded video ;)

davydovanton:
folks, let's merge this amazing PR? :)

mbajur:
Just let me bundle the assets in to the project. I was just about to do that so it should be done in 20 minutes :)

mbajur:
Ok, this thing is ready to be merged.

mbajur:
Ok, this thing is ready to be merged. I also created a demo on heroku:

https://hanami-docs-search.herokuapp.com

davydovanton:
Hey, I found the strange bug. I'm not sure that you can (and should) fix it, but I think it will be helpful to show this

42nlmueo46

mbajur:
Ok, i forgot to escape html in the excerpts. Fixed and deployed now :)

cllns:
:+1: this will be great :)

PR #332, opened 2017-05-20
Added selection of `foreign key`

ilyario:


AlfonsoUceda:
@ilyario thanks for this PR but I agree with @davydovanton, we don't need this change

ilyario:
OK.
I thought that it was necessary to allocate a function, as in other paragraphs.

cllns:
I'm actually in favor of this, if you change it to `foreign_key` (with the underscore).

`primary_key` has the code blocks around it, in the previous paragraph and `#column` has it in the next paragraph. We should be consistent :)

While you're at it , can you remove the leading `#` from `#column`? 

Are you OK with that, @AlfonsoUceda & @davydovanton?

AlfonsoUceda:
@cllns :+1:

davydovanton:
@cllns 👍 

@ilyario  could you fix it the small issue, and after that, I'll merge this changes? thank you!

ilyario:
@davydovanton here are not edited locally, this recreate PR
see https://github.com/hanami/hanami.github.io/pull/348

davydovanton:
@ilyario okay I'll close this PR

PR #333, opened 2017-05-21
Add information about creating batch records

davydovanton:
/cc @hanami/core

### DIFF
--- a/source/guides/1.0/models/repositories.md
+++ b/source/guides/1.0/models/repositories.md
@@ -63,6 +63,27 @@ repository.find(book.id)
   # => nil
 ```
 
+## Saving more than one record
+
+For saving bunch of records you need to create `#create_many` command:
+
+```ruby
+class BookRepository
+  def create_many(list)
+    command(:create, books, result: :many).call(list)
+  end
+end
+```
+
+Now you can use `#create_many` a bulk creation method:
+
+```ruby
+repository = BookRepository.new
+
+repository.create_many([{ title: "Hanami" }, { title: "Ruby"}])
+# => [#<Book:0x007f95cbd8b7c0 @attributes={...}>, #<Book:0x007f95cbd8b7c0 @attributes={...}>]
+```
+
 ## Private Queries
 
 **All the queries are private**.
--- a/source/guides/1.0/models/repositories.md
+++ b/source/guides/1.0/models/repositories.md
@@ -65,7 +65,7 @@ repository.find(book.id)
 
 ## Saving more than one record
 
-For saving bunch of records you need to create `#create_many` command:
+For saving bunch of records you need to define a `#create_many` command:
 
 ```ruby
 class BookRepository
@@ -75,13 +75,11 @@ class BookRepository
 end
 ```
 
-Now you can use `#create_many` a bulk creation method:
+Now you can use `#create_many` as a bulk creation method:
 
 ```ruby
-repository = BookRepository.new
-
-repository.create_many([{ title: "Hanami" }, { title: "Ruby"}])
-# => [#<Book:0x007f95cbd8b7c0 @attributes={...}>, #<Book:0x007f95cbd8b7c0 @attributes={...}>]
+BookRepository.new.create_many([{ title: "Hanami" }, { title: "Ruby"}])
+# => [#<Book:0x007f95cbd8b7c0 @attributes={...}>, #<Book:0x007ff4e20e6f80 @attributes={...}>]
 ```
 
 ## Private Queries


davydovanton:
or we can add something like this:
```ruby
class BookRepository
  def create_many(list)
    command(:create, books).call(list)
  end
end
```

source/guides/models/repositories.md
```diff
@@ -50,6 +50,9 @@ repository = BookRepository.new
 book = repository.create(title: "Hanami")
   # => #<Book:0x007f95cbd8b7c0 @attributes={:id=>1, :title=>"Hanami", :created_at=>2016-11-13 16:02:37 UTC, :updated_at=>2016-11-13 16:02:37 UTC}>
 
+# you can create more than one record in transaction
+repository.create([{ title: "Hanami" }, { title: "Ruby"}])
```
AlfonsoUceda: We can add too the result of this

davydovanton:
@AlfonsoUceda I changed PR and added information how to create `#create_many` method with ROM command

@hanami/core WDYT?

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
   # => nil
 ```
 
+## Saving more than one record
+
+For saving butch of records you need to create `#create_many` command:
```
b264: `s/butch/a bunch/`

AlfonsoUceda: @davydovanton can you amend this?

AlfonsoUceda:
@davydovanton one question, do you know if the result of this query is one insert or several inserts?

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
...
+
+```ruby
+class BookRepository
+  def create_many(list)
+    command(:create, books, result: :many).call(list)
+  end
+end
+```
+
+And after that you can use it as a `#create` method:
```
TiteiKo: I find this sentence misleading, it may lead to people thinking the `#create` method can now create many records instead of understanding that there's a new `#create_many` method

It also can't be used as the `#create` method, as the parameters passed are really different...

I'd rephrase to `Now you can use `#create_many` as a bulk creation method:`

davydovanton:
@AlfonsoUceda if I remember right it's only one command (`INSERT`) [source](https://github.com/rom-rb/rom-sql/blob/9ea3f06bd61a8db87ea7b71ebb9fc2bb75767f58/lib/rom/sql/relation/writing.rb#L43)

@TiteiKo great idea, I like it 👍
Thank you for fixing my bad English :D

davydovanton:
Updated text, thank you all for review ❤️

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
...
+
+```ruby
+class BookRepository
+  def create_many(list)
+    command(:create, books, result: :many).call(list)
+  end
+end
+```
+
+Now you can use `#create_manyas` a bulk creation method:
```
TiteiKo: Typo here `#create_manyas`

davydovanton: done 👍

davydovanton:
@AlfonsoUceda @TiteiKo @b264 any other issues or I can merge this?

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
   # => nil
 ```
 
+## Saving more than one record
+
+For saving bunch of records you need to create `#create_many` command:
```
cllns: How about:

For saving bunch of records you need to **define a**`#create_many` command:

create #create_many looks a little strange to me :)

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
...
+
+```ruby
+class BookRepository
+  def create_many(list)
+    command(:create, books, result: :many).call(list)
+  end
+end
+```
+
+Now you can use `#create_many` a bulk creation method:
```
cllns: Now you can use `#create_many` **as** a bulk creation method:

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
...
+
+```ruby
+repository = BookRepository.new
+
+repository.create_many([{ title: "Hanami" }, { title: "Ruby"}])
+# => [#<Book:0x007f95cbd8b7c0 @attributes={...}>, #<Book:0x007f95cbd8b7c0 @attributes={...}>]
```
cllns: pedantic but I don't think the Books would have the same object id ;)

source/guides/models/repositories.md
```diff
@@ -62,6 +62,27 @@ repository.find(book.id)
...
+  end
+end
+```
+
+Now you can use `#create_many` a bulk creation method:
+
+```ruby
+repository = BookRepository.new
+
+repository.create_many([{ title: "Hanami" }, { title: "Ruby"}])
```
cllns: Since it's just one command, should we do `BookRepository.new.create_many`, instead of defining a local variable?

AlfonsoUceda:
@davydovanton can you take care of this? :P

davydovanton:
Hey, yes. I'll do it after Japan :)

AlfonsoUceda:
Merging because I've fixed suggestions.

davydovanton:
thank you ❤️

PR #336, opened 2017-05-26
Fix Escape Helpers link in Guides HTML5 Helpers

cveneziani:
The link to Escape Helpers page is broken on HTML5 helpers page.
See http://hanamirb.org/guides/helpers/html5/

- Broken link: http://hanamirb.org/guides/helpers/html5/%60/guides/helpers/escape%60 (404 page)
- Correct link: http://hanamirb.org/guides/helpers/escape/


cllns:
Great find, thank you so much!! 🌸

PR #337, opened 2017-05-26
Fix Guides - Request and Response page

cveneziani:
Small typo here, we should use `accept?` method instead of `accepts?` method: there is no `s`.

I'm new to the framework so here I guess it refers to this method 👉 https://github.com/hanami/controller/blob/master/lib/hanami/action/mime.rb#L488-L492


cveneziani:
Oh nice, thanks for checking :)

PR #338, opened 2017-05-30
Update create-table.md

jarosluv:
With Ruby 2.4 and Hanami 1.0.0, `Bignum` becomes `integer` in a migration SQL. From now we need to use `Bignum` as a symbol `:Bignum`.


jarosluv:
I just found this code in Sequel sources: 

```
# Use custom Bignum method to use :Bignum instead of Bignum class, to work
# correctly in cases where Bignum is the same as Integer.
def Bignum(name, opts=OPTS)
  column(name, :Bignum, opts)
end
```

This must mean that we can use plain Bignum for our purposes. May be we need relevant fallback for Hanami code?

davydovanton:
@jarosluv thanks for you work 💚

PR #342, opened 2017-06-12
Update getting started guide

damireh:
I was going through the guide and this part surprised me

> As you can see, in this case we simply hard-code the error message "is required", but you could inspect the error and customise your message for the specific validation that failed

because the guide didn't hardcode "is required" it looped over `param.errors` and showed proper error messages.

Wasnt sure what to replace it with, so I copy pasted the "run your tests again" message from the section above since the code snippet below is running the tests.


davydovanton:
@hanami/core WDYT?

davydovanton:
@damireh thanks for your work! 💚

PR #346, opened 2017-06-20
Fix custom repository query example.

beauby:


source/guides/models/repositories.md
```diff
@@ -95,6 +95,7 @@ class BookRepository < Hanami::Repository
       .where(author_id: author.id)
       .order(:published_at)
       .limit(limit)
+      .to_a
   end
 end
```
davydovanton: So, if we add `#to_a` here we need to add it [here](https://github.com/hanami/hanami.github.io/pull/346/files#diff-d35c518e353ad9c912f2e43ed7db26f6R73). WDYT?

davydovanton: And we need this method here? Maybe is better to add a description why we call `#to_a`?

beauby: Regarding your first remark, it is not clear as the code snippet is hypothetical anyways. I do not feel strongly about that though so if you think it's better to add `.to_a` there as well then I'll do it.

Regarding your second remark, we need to materialize the query in order to be consistent with the above remark ("The caller has an intimate knowledge of the internal mechanisms of the Repository."). If we do not call `#to_a`, then the method returns some ROM object instead of a collection of entities.

davydovanton: 1. yep, I want to see equal style in this places (with or without `#to_a`)
2. I think we should add simple description about `#to_a` method before all this

AlfonsoUceda:
@beauby is there a chance to fix this?

PR #349, opened 2017-06-23
Fix formatting for 'hanami routes' command output

davydovanton:
### Before
screenshot 2017-06-23 10 27 27

### After
screenshot 2017-06-23 10 27 18

/cc @hanami/core


oana-sipos:
Nice one, @davydovanton!

PR #351, opened 2017-07-03
Structural improvements to overview.md

Nikamura:
It makes more sense for *Rack* description to be before *Actions*.

### DIFF
--- a/source/guides/routing/overview.md
+++ b/source/guides/routing/overview.md
@@ -36,23 +36,6 @@ trace   '/hello', to: endpoint
 options '/hello', to: endpoint
 ```
 
-## Actions
-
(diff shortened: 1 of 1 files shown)


davydovanton:
@hanami/core WDYT? I don't see any benefits here but maybe you will see

Nikamura:
at least how the text is formatted, I was confused as you introduced Rack.

> Full Rack integration is great, but the most common endpoint that we'll use in our web applications...

even though it has not been introduced and only follows after *actions*. 
First reading about Rack then about actions makes more sense.

davydovanton:
okay, it looks convincing. Thanks!

davydovanton:
thanks for contribution!

PR #355, opened 2017-07-24
Add missing example with option args

davydovanton:


davydovanton:
cc @hanami/core

davydovanton:
@AlfonsoUceda 💚

PR #356, opened 2017-07-24
Separate guides for each version

davydovanton:
screenshot 2017-08-02 19 24 32

screenshot 2017-08-02 19 24 58

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'middleman-syntax',                '~> 2.0'
 gem 'middleman-deploy',                '~> 1.0'
 gem 'middleman-search_engine_sitemap', '~> 1.3'
 gem 'middleman-search'
+gem 'middleman-breadcrumbs'
 gem 'redcarpet',                       '~> 3.2'
 gem 'httparty'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -70,6 +70,8 @@ GEM
       addressable (~> 2.3.5)
       middleman-core (~> 3.2)
       tzinfo (>= 0.3.0)
+    middleman-breadcrumbs (0.3.3)
+      middleman (>= 3.3.5)
     middleman-core (3.4.1)
       activesupport (~> 4.1)
       bundler (~> 1.1)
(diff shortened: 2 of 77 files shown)


cllns:
Thanks for working on this! It's really important.

I don't like the idea of having many version of each guide in the same repository context. It seems like a mess to maintain, particularly since most people will only be concerned with the most recent stable version. Especially with all the link changes.

Could we, somehow, leverage git tags (or branches) to maintain different versions? I don't know. 

Do we know how Rails maintains old guides?

davydovanton:
@cllns yep, this solution isn't the best but it was most easily and faster way to do this prototype. I think we can research and change this logic in future. WDYT?

oana-sipos:
@cllns I don't know how they handle it, but what about asking somebody in the Rails team on Twitter?

davydovanton:
@oana-sipos you can find all information here :)
https://github.com/rails/rails/tree/master/guides/rails_guides

jodosha (commented):
@davydovanton Thanks for working on this. 💯 

---

I've noticed that we don't have any clear indication that we're in the Guides section:

screen shot 2017-07-26 at 19 19 48

Possible ideas (not mutually exclusive)

  * Background highlight for the "Guides" voice in the menu
  * Breadcrumbs (eg. `Guides > 1.0 > Actions > Overview`)

---

By looking at the screenshot above, there isn't any clear indication of the current version of Hanami.

---

Only if I switch to an "old" version I see the version:

screen shot 2017-07-26 at 19 24 56

But again, it's better to **always** have the Hanami version clear at the top of the page.

---

The switch for the Hanami versions doesn't really belongs to the right menu:

screen shot 2017-07-26 at 19 26 38

jodosha:
@cllns 

> I don't like the idea of having many versions of each guide in the same repository context. It seems like a mess to maintain, particularly since most people will only be concerned with the most recent stable version. Especially with all the link changes.
> 
> Could we, somehow, leverage git tags (or branches) to maintain different versions?

Yes, it can be a burden. But what's potentially each article for the same subject (eg routing) can different from the same article for a different version.

The idea is to have all the versions available online. How would you handle this? We're open to suggestions. 👍

PR #359, opened 2017-08-05
Added hlegius and Creditas as donators

hlegius:
Hallo all!

I'm adding Creditas (with logo) and myself (link only) as donators on Donation page.

I was wondering how could be the _best_ way to place Creditas' logo there and did using `<p>` inside `<div>` - not sure if it's the best way™ to do so.

This is how it is now:

screen shot 2017-08-05 at 7 15 44 pm

Thoughts? 💯


jodosha:
@hlegius Thank you and Creditas for the support. 💚

PR #360, opened 2017-08-06
Update entities.md

Shkrt:
Typo :)


oana-sipos:
Thanks for the correction @Shkrt!

PR #376, opened 2017-09-23
Update link to migration generator

cveneziani:
On [Migration Overview](http://hanamirb.org/guides/1.0/migrations/overview/) page, just make sure that the link to the **migration generator** goes directly at the `#migrations` anchor, [here](http://hanamirb.org/guides/1.0/command-line/generators/#migrations).

I made the change for both `1.0` and `head` versions.


cllns:
:+1: Thank you! 🌸

PR #379, opened 2017-10-01
Blogpost documentation challenge

marionschleifer:
This blogpost is about the documentation challenge that we have in October, where we award contributions to our documentation with Hanami gadgets.


oana-sipos:
Just my 2 cents, but wouldn't it be better to call it an action alongside [Hacktoberfest](https://hacktoberfest.digitalocean.com/)? Or we prefer not to associate ourselves with them?

marionschleifer:
@oana-sipos I also thought about that, but our campaign will most likely be extended for the coming months, and if we associate ourselves with the Hacktoberfest, everybody thinks that it's limited.

PR #382, opened 2017-10-02
Mention PostgreSQL types in Migration data types page

cveneziani:
When looking for wich data type I can use in a migration, I know that it must be somewhere in the Migrations section. And yes, I find the information here in the [Migrations create table](http://hanamirb.org/guides/1.0/migrations/create-table/) page.

But I keep forgetting that there are extra data types for PostgreSQL that are mentionned elsewhere, here in [Models PostgreSQL](http://hanamirb.org/guides/1.0/models/postgresql/).

So I've added a mention to this page on **Migrations create** table page.

BTW does it make sense to have [PostgreSQL](http://hanamirb.org/guides/1.0/models/postgresql/) page inside the **Models** category? From what I see, it's mostly about PostgreSQL migrations. What do you think?

PS: the default purple link didn't look nice inside the convention paragraph. I turned it into white color but I'm not sure how CSS is structured. So I've added a property at the end of `source/stylesheets/guides.css` file.

### DIFF
--- a/source/guides/1.0/migrations/create-table.md
+++ b/source/guides/1.0/migrations/create-table.md
@@ -50,7 +50,10 @@ It supports the following options:
   * `:primary_key` (make the column primary key for the table)
   * `:unique` (add a uniqueness constraint for the column)
 
-
+<p class="convention">
+  Note that Hanami natively supports <strong>PostgreSQL data types</strong>.
+  Learn more about them in the <a href="/guides/1.0/models/postgresql/">dedicated article</a>.
+</p>
 
 #### Primary Key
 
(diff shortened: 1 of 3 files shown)


davydovanton:
@cveneziani could you show screenshot with `convention` block? Thanks! 💙

cveneziani:
Sure, here it is 👇 

pg-data-types

davydovanton:
@cveneziani wow, awesome 👏

AlfonsoUceda:
thanks @cveneziani it look nice. You are right about Postgresql model page, maybe we should unify that in migrations or create table page, for now I'm going to merge this PR.

cveneziani:
I guess that it would make sense to just move this PostgreSQL page within the migrations category.

Thanks for the merge.

Issue #309, opened 2017-04-06
Add information about joins in hanami/rom


davydovanton:
I think it's an important theme for guides. That's why I suggest to add this. We can use this SO question for example:
https://stackoverflow.com/questions/43080678/join-query-in-hanami-model

alvir:
I was confused with query documentation in guides. I think It will be helpful to add info about rom-sql and Sequel.

mereghost:
Some guides would be nice, specially to the features where we meddle with rom (associations and the like).

Covering all the queries would make our docs coupled to `rom`'s docs, so it would be better to point users to `rom` docs.

cveneziani:
What about this issue now, since Pull Request #386 has been merged?

davydovanton:
@cveneziani hey! you can take my old PR and update it if you want https://github.com/hanami/hanami.github.io/pull/344 :)

cveneziani:
Well, as @mereghost mentioned. Shall we copy the doc from ROM or just point out to ROM doc? And if we copy, where shall we stop?

What's you point of view?

PR #386, opened 2017-10-03
Add mention to ROM and Sequel in Repositories guide

cveneziani:
I'm often struggling to create rather "complex" SQL queries (eg. joins, order desc, average, etc.) with Hanami repository. And I often end up with a mix of ROM and exceptionally Sequel syntax (eg. when having to qualify a column).

I think it could be nice to provide links to ROM and Sequel documentations in the [Repository](http://hanamirb.org/guides/1.0/models/repositories/#private-queries) guide as mentioned here https://github.com/hanami/hanami.github.io/issues/309#issuecomment-300313160

Here is a screenshot of the block I've added 👇 

mention-rom-n-sequel

### DIFF
--- a/source/guides/1.0/models/repositories.md
+++ b/source/guides/1.0/models/repositories.md
@@ -131,6 +131,19 @@ This is a **huge improvement**, because:
 
   * If we change the storage, the callers aren't affected.
 
+<div class="convention">
+  <p>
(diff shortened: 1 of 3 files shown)


cveneziani:
Happy to help 😃

PR #408, opened 2017-10-30
Assets overview grammar

philnash:
I was reading through this page and found a few places that could be tidied up to make more sense. I think I've retained the original meanings, just think it reads better this way.

Hope this is helpful!


oana-sipos:
Thanks @philnash for your contribution :+1:  :white_flower:

philnash:
Happy to help out! The guides are so well thought through, it's my pleasure to help them out in just a little way.

PR #414, opened 2017-11-02
Make clarification about the test/spec defaultness

apohllo:
I'm not sure if this is true, but it seems from the context.


oana-sipos:
Thanks @apohllo for the clarification made :100:

marionschleifer:
@apohllo, thank you very much for your contribution 🌷 Are interested in receiving some Hanami stickers? If yes, could you send me your postal address via email (marion.schleifer@gmail.com) 😃

PR #422, opened 2017-12-09
Accessing values exposed by the controller

apohllo:
I haven't found any info about the fact, that you can access the values from the controller in the view. This is particularly important for views that want to wrap these values or convert them to other representation (e.g. JSON).

### DIFF
--- a/source/guides/1.1/views/basic-usage.md
+++ b/source/guides/1.1/views/basic-usage.md
@@ -103,8 +103,6 @@ module Web::Views::Dashboard
 end
 ```
 
-
-
 ## Custom Rendering
 
 Hanami performs rendering by calling `#render` on a view and it expects a string in return.
--- a/source/guides/1.1/views/basic-usage.md
+++ b/source/guides/1.1/views/basic-usage.md
@@ -86,6 +86,25 @@ end
 Rendering context for a template is made of view methods and exposures.
 </p>
 
+The objects exposed in the controller action are available in the corresponding view. So the values 
+can also be modified, wrapped or reused in some other way. Assuming that the `title` is exposed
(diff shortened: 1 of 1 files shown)


apohllo:
Recently I've found that this information is in fact present in the Actions guide. Yet I think it is vital to repeat it in the View guide, since this seems to be the primary (or at least secondary) place users will look for such an information.

source/guides/1.1/views/basic-usage.md
```diff
@@ -86,6 +86,25 @@ end
...
+    
+    def full_title
+      "The title: " + title
+     end
+  end
+end
+```
+
+
+
```
davydovanton: could you cut unnecessary lines? :)

source/guides/1.1/views/basic-usage.md
```diff
@@ -86,6 +86,25 @@ end
 Rendering context for a template is made of view methods and exposures.
 </p>
 
+The objects exposed in the controller action are available in the corresponding view. So the values 
+can also be modified, wrapped or reused in some other way. Assuming that the `title` is exposed
+in the action, it can be accessed as follows:
+
+```ruby
+# apps/web/views/dashboard/index.rb
+module Web::Views::Dashboard
```
davydovanton: WDYT maybe it will be good to add controller class with `title` expose?

apohllo: It is defined two blocks of code before (in the same section). I don't think it would be helpful to write it once again.

davydovanton: great point, thanks

apohllo:
Your welcome!

PR #425, opened 2017-12-22
No params logging howto

graywolf:
Add info how to completely disable logging of params of requests.


AlfonsoUceda:
@graywolf thanks for this PR, can you explain why addition? I'd like to know what is the use case to disable logging with a custom formatter. Thanks!

graywolf:
Mainly to disable logging of submitted data for security and compliance reasons. While I can manually supress specific field name, this is best way (afaict) do disable it completely. So I thought it would be nice to have it documented so people don't have to go digging into source code/gitter chat to figure this one out like I did.

PR #428, opened 2018-01-05
Add example for accessing request headers

DSIW:


davydovanton:
@DSIW and also, could you update this for `1.0` version too? Thanks ❤️

DSIW:
Maybe it would be good to add this somewhere in the `actions/parameters.md` too, because the headers are also accessible by `params.env['...']`. Do I understand it correctly that the access by `params` should be preferred for better performance?! That should be emphasized, too. Do you know where it could be a good place?

PR #431, opened 2018-01-30
Expands the info about legacy database support

mereghost:
Adding some info on how to deal with legacy databases, the caveats of mapping and working with just a subset of the schema.

I'm open to suggestions.

### DIFF
--- a/source/guides/1.1/repositories/overview.md
+++ b/source/guides/1.1/repositories/overview.md
@@ -246,7 +246,7 @@ class OperatorRepository < Hanami::Repository
   end
 
   def by_name(name)
-    t_operator.where(s_name: name)
+    root.where(s_name: name)
   end
 end
 ```
--- a/source/guides/1.1/repositories/overview.md
+++ b/source/guides/1.1/repositories/overview.md
@@ -217,14 +217,14 @@ operator = OperatorRepository.new.create(name: "Jane")
   # => #<Operator:0x007f8e43cbcea0 @attributes={:id=>1, :name=>"Jane"}>
 ```
 
-There's a few caveats tho (as always with legacy databases):
+There are few caveats though (as there always are, with legacy databases):
 
-Note that all attributes need to be mapped or you will get an error when booting up the application that will look like this:
+Note that **all** attributes need to be mapped or you will get an error when booting up the application that will look like this:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


source/guides/1.1/repositories/overview.md
```diff
@@ -218,6 +217,72 @@ operator = OperatorRepository.new.create(name: "Jane")
...
+    attribute :s_name, Hanami::Model::Sql::Types::String.meta(alias: :name)
+  end
+  
+  mapping do
+    attribute :id,   from: :operator_id
+    attribute :name, from: :s_name
+  end
+  
+  def by_name(name)
+    t_operator.where(name: name)
```
davydovanton: can we use `#root` here?

source/guides/1.1/repositories/overview.md
```diff
@@ -218,6 +217,72 @@ operator = OperatorRepository.new.create(name: "Jane")
...
+```
+
+If you want to use the more semantic names (or with just a subset of attributes) you'll need to define a schema and a mapping:
+
+```ruby
+class OperatorRepository < Hanami::Repository
+  self.relation = :t_operator
+  
+  schema do
+    attribute :operator_id, Hanami::Model::Sql::Types::Int.meta(primary_key: true, alias: :id)
```
davydovanton: just idea, but maybe we can use more friendly types here?

AlfonsoUceda:
@mereghost can you ask to @davydovanton I think after that we can merge it.

mereghost:
@AlfonsoUceda Sure. I was away-ish the last few days. I'll address his points and update this.

cllns:
Cleaned up some of the language, but I'm not comfortable enough with this material to address @davydovanton's points :)

source/guides/1.1/repositories/overview.md
mereghost: sure.

source/guides/1.1/repositories/overview.md
mereghost: While we probably could use `Dry::Types` not sure if it wouldn't lead to confusion as the end user might think something along the likes of "oh I need yet another lib to do this".

mereghost:
@cllns Thanks for the help. =)

PR #438, opened 2018-02-21
Add notes about removing from `spec` dir

Pistos:
And also reformat the listing.


cllns:
Thanks for this @Pistos!

I added a commit that changes the PR significantly, since I think we should focus this section on only removing `hanami-model`, and leave removing Mailers to the reader. I know that was already there before, but what do you think of my changes in general?

Pistos:
Thank you for your attention to this PR.

kaikuchn:
looks good to me!

cllns:
Thanks @Pistos & @kaikuchn!

PR #443, opened 2018-03-01
Convert really long let blocks to multi-line variants in GS guide

radar:
This makes the code fit better on the page. These `let` blocks were wrapping on my laptop screen, which made it a little hard to parse the code.

With multi-line let blocks, the code doesn't wrap and it looks much better.

### DIFF
--- a/source/guides/getting-started.md
+++ b/source/guides/getting-started.md
@@ -513,8 +513,16 @@ require_relative '../../../../apps/web/views/books/index'
 
 describe Web::Views::Books::Index do
   let(:exposures) { Hash[books: []] }
-  let(:template)  { Hanami::View::Template.new('apps/web/templates/books/index.html.erb') }
-  let(:view)      { Web::Views::Books::Index.new(template, exposures) }
+  let(:template)  do
+    Hanami::View::Template.new(
+      'apps/web/templates/books/index.html.erb'
+    )
+  end
+
+  let(:view) do
+    Web::Views::Books::Index.new(template, exposures)
+  end
+
   let(:rendered)  { view.render }
 
   it 'exposes #books' do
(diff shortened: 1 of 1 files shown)


cllns:
Thanks! Both for this PR and for going through the Getting Started with `1.2.0.beta1` so soon :D 

I agree the single-lines look bad in a browser. (In my Hanami projects, I change all my > 80 character line `let`s into multi-line blocks, since my text editor draws a line at 80 characters).

I don't expect it to be a _huge_ deal but this causes a discrepancy between what the generator does (single-line blocks) and what's on the Getting Started guide (multi-line, in these cases). I'd like to keep them identical if possible, but if we can't, most readers will know that the blocks are equivalent.

We could make the generator smarter to break into multi-line blocks at 80 character lines, but that doesn't quite fix this problem, since these lines are broken at fewer than 80 characters.

I wonder if we could, instead, fix this in the CSS? I'm not sure how though. There are a few breakpoints that could be simplified, which would add some room for wider code blocks, below the navigation.

Thoughts?

radar:
Well yeah, you could fix it in the CSS by changing it to overflow: auto or overflow: scroll (I think). Then people would just need to scroll across in the code examples.

We could also fix the generators to prefer multi-line do blocks but I think that’s more work :)

apohllo:
As far as I am concerned I strongly prefer single-line `let` blocks.  The reason is that the `let`s form a concise list of variables which allows to quickly determine the exact context of the test. On the other hand - in `do` blocks the variables are interleaved with the code to compute their values, which is a cognitive overhead. Thus whenever I have to write multi-line `do` blocks I always put them at the end in order keep the context definition easy to grasp.

jodosha:
@radar Thanks for this PR. Guides has been removed from this repository. See #478

PR #446, opened 2018-03-17
Guides TOC as accordion menu

mbajur:
Hello!

Guides TOC menu is a really long one and it's very hard to find where we're currently at when browsing the guides (we need to scroll it all over and look for a string [guide page title] as it seems that `active` link class is not working on live hanami guides website) so i decided to propose a change of way TOC is rendered. My suggestion is an accordion menu, it looks more or less like that:

image

It's way easier to follow a guide this way as all the links from current category are visible on the screen and you don't need t scroll all over the place to find them.


mbajur:
Ok, i decided to check why the `active` class is not working on live website and it turned out it was caused by the `guide_url` helper not adding a trailing slash to the urls.

Fix is in that PR :)

AlfonsoUceda:
@mbajur thanks for this PR, merging it!

PR #448, opened 2018-03-18
Update search to only show newest version

siaw23:


source/guides/1.1/getting-started.md
```diff
@@ -150,7 +149,7 @@ describe 'Visit home' do
   it 'is successful' do
     visit '/'
 
-    page.body.must_include('Bookshelf')
+    page.must_include('Bookshelf')
```
AlfonsoUceda: this is not true, we need to search in body to make this test passes

AlfonsoUceda:
@siaw23 this PR is not needed anymore because we already support Algoria, thanks for your work anyway ;)

PR #453, opened 2018-03-31
Improve readme

siaw23:


siaw23:
hi @cllns can you please check this for me? and one question; i mentioned before the test results of what is in the guides don't match what actually happens on users' computers. not sure if i should investigate this because i saw some chats about switching over to RSpec.

source/guides/1.0/getting-started.md
```diff
@@ -474,7 +474,7 @@ We can use migrations to apply changes to our database schema.
 
 ### Displaying Dynamic Data
 
-With our new experience modelling data, we can get to work displaying dynamic data on our book listing page.
+With our new experience modeling data, we can get to work displaying dynamic data on our book listing page.
```
siaw23: here i realised the guide is using american english, so using one type of english everywhere only makes sense.

source/guides/1.0/getting-started.md
```diff
@@ -1078,7 +1078,11 @@ redirect_to routes.books_path
 
 **Congratulations for completing your first Hanami project!**
 
-Let's review what we've done: we've traced requests through Hanami's major frameworks to understand how they relate to each other; we've seen how we can model our domain using entities and repositories; we've seen solutions for building forms, maintaining our database schema, and validating user input.
+Let's review what we've done:
+
+* We've traced requests through Hanami's major frameworks to understand how they relate to each other.
+* We've seen how we can model our domain using entities and repositories.
+* We've seen solutions for building forms, maintaining our database schema and validating user input.
```
siaw23: nice to make it a lis of what has been accomplished, easier to read that way i think,

source/guides/1.0/getting-started.md
```diff
@@ -410,7 +410,7 @@ end
 ```
 
 Hanami provides a DSL to describe changes to our database schema. You can read more
-about how migrations work in the [migrations' guide](/guides/1.0/migrations/overview).
+about how migrations work in the [Migrations Guide](/guides/1.0/migrations/overview).
```
siaw23: here "migrations guide" as an entity, the migrations themselves don't have "guides" so it looks a bit funny to say "migrations' guide" :D :D or even "migration's guide" for one migration.

source/guides/1.0/getting-started.md
```diff
@@ -256,7 +256,7 @@ Finished in 0.011854s, 84.3600 runs/s, 168.7200 assertions/s.
 Let's use our new knowledge about the major Hanami components to add a new action.
 The purpose of our Bookshelf project is to manage books.
 
-We'll store books in our database and let the user manage them with our project.
+We'll store books in our database and let the user manage them with our application.
```
davydovanton: in hanami we use project for full repository and application for part of the project in `apps/` folder. In this case, `project` better raise than `application`

siaw23: ah i understand. but here 'application' as in software application that enables us to do something. so example Word is an application that allows us to write documents or Bookshelf in this case is an application that helps us store books. 'application' here doesn't mean '/apps' folder, it means the [program/software/web app](https://en.wikipedia.org/wiki/Application_software) we're building. but i can change it back if you're still not convinced.  @davydovanton

AlfonsoUceda:
thanks @siaw23 merging!

PR #462, opened 2018-04-18
Make the behaviour of exposures more clear

kj:
I found this a bit confusing as the guide currently makes it sound a bit like it works by passing instance variables to the view, but actually it works with any method of the given name and only creates an `attr_reader` if a method doesn't already exist. Does this make sense? If so, should it also be applied to the 1.1 and 1.0 articles?


jodosha:
@kj Thank you very much. 👍

PR #465, opened 2018-06-29
Use travis's first-class deploy to GH pages

Morozzzko:
Travis supports deploy to GH pages as a [first-class feature](https://docs.travis-ci.com/user/deployment/pages/).

So I figured it would be nice to use the feature instead of hacky `before_script`.

As a plus, it will allow us to build pull requests to see if they fail.


GustavoCaso:
Thank you so much for this contribution. ❤️

PR #466, opened 2018-06-30
Add PPE analytics to the list of companies using Hanami in production

inouire:


inouire:
You're welcome, thanks for the merge.

jodosha:
@inouire Thanks for using Hanami. 🌸

PR #483, opened 2018-12-14
Remove pointless CSS rule

jacobherrington:
I noticed this gap and can't see a reason for this CSS rule

Before:
image

After:
image


AlfonsoUceda:
Thanks @jacobherrington, it doesn't make any sense.

PR #497, opened 2020-01-17
Add Ascenda Loyalty to list of companies using Hanami

Drenmi:
Firstly, thank you so much for all your hard work on making Hanami possible.

At Ascenda, we have been using Hanami in production for a couple years already, and it has helped us write better factored apps, and keeps us growing as developers.

We would love to be part of the list of companies who trust Hanami.

💚💜💛❤️💙


davydovanton:
Hey @Drenmi! Happy to see your company here and thank you for this PR! ❤️

PR #499, opened 2020-01-17
Use html entity for extract

### LINKED ISSUE #498: Atom feed is broken
When visiting https://hanamirb.org/atom.xml I get the following error:

```
XML Parsing Error: not well-formed
Location: https://hanamirb.org/atom.xml
Line Number 84, Column 79:
         Completely rewritten validation library, compatible with latest DRY & ROM versions. Use it today with Hanami 1!
------------------------------------------------------------------------------^
```

mauromorales:
fixes #498

### DIFF
--- a/source/atom.xml.erb
+++ b/source/atom.xml.erb
@@ -21,7 +21,7 @@
          <name><%= article_author(article) %></name>
        </author>
        <summary>
-         <%= article_summary(article) %>
+         <%= ERB::Util.html_escape(article_summary(article)) %>
        </summary>
        <content type="html"><![CDATA[<%= article.body %>]]></content>
      </entry>


cllns:
Thanks @mauromorales! 🌸 

Do you think there's a change we could make at a different level, to either fix this automatically (by escaping special characters) or at least trigger some kind of error, to let us know we need to manually use the html entity?

mauromorales:
@cllns good call, I've wrapped the extract in the atom layout to encode HTML entities. I decided to only do it there and not at the helper function since I'm only aware it's breaking for RSS. What do you think?

cllns:
Thanks @mauromorales for the quick turnaround! I just reproduced this issue and see that the changes in this PR fixes it :)

There is an escaping method in `hanami-utils` at [`Hanami::Utils::Escape.html(...)`](`https://github.com/hanami/utils/blob/master/lib/hanami/utils/escape.rb#L420`) but that's not used in this project and it seems like it's not worth bringing in the dependency for something that the standard library includes anyway (even if there's a performance or security benefit, since this is code that's been checked in through GitHub, and deployed statically).

cllns:
Fix is live on production: https://hanamirb.org/atom.xml 🎆

mauromorales:
@cllns thank you for the feedback!

PR #500, opened 2020-01-21
Fix grammar on homepage.

mehagar:


cllns:
Thanks for the suggestion. I don't think it's necessary though. Articles are optional before plural nouns.

If you want a contribution for changing this line, I'd happily merge adding an `and` before `automatic` (while retaining the comma after `headers`)  :)

mehagar:
I agree it's not necessary, it's certainly awkward though to read and not idiomatic english. See [this answer](https://english.stackexchange.com/questions/114467/omitting-article-the-in-front-of-plural-nouns).

PR #506, opened 2020-02-28
Improve blog reading

AlfonsoUceda:
I've added some css to make better reading for small and large devices. This blog isn't responsive but at least make it looks better.

Also I've updated and removed some dependencies.

 ## MOBILE 
Screenshot_2020-02-28 Introducing Hanami API(1)
Screenshot_2020-02-28 Hanami Blog

## DESKTOP
Screenshot_2020-02-28 Introducing Hanami API
Screenshot_2020-02-28 Hanami Blog(1)


AlfonsoUceda:
@jodosha I'll try. That widget is in the screenshot because of the tool I used for taking the whole screenshot but I will try it ;)

Issue #508, opened 2020-07-29
Hanami security page is broken


dgollahon:
I was trying to submit a report of a possible vulnerability with an older version of Hanami that I couldn't find documented anywhere, but the security link on the homepage made that hard to do. It prompted me to download a text file which seems to be a random chunk of HTML instead of taking me to a real page.

I found `security@hanamirb.org` in that file but when I sent an email to that address it bounced.

At minimum this seems like a bug with the website and it would be good to have a clear/easily accessible security policy and contact point. If `security@hanamirb.org` (or an alternative) is not going to be maintained, there should at least be a notice saying that this project does not care to accept vulnerability reports.

cllns:
Ooof thanks for reporting this and responsibly disclosing a potential vulnerability. We'll fix this issue with the security page being broken and the email address bouncing. We definitely take these seriously, we're just all volunteers & without a financial sponsor, so some things fall through the cracks :(

In the meantime, if you want to send the report to me, I can look into it and circulate it with the rest of the core team. `sean at this-github-handle dot com`

dgollahon:
Sure thing.

I sent you an email with the potential issue.

I know it's a lot of work to run a project like this, thanks for helping improve the page/response here.

cllns:
Received. Thank you! 

Let's leave this issue open until we can get a PR merged fixing the security page being broken. Also working on fixing the email address bouncing too.

cllns:
Closed with #509

PR #509, opened 2020-08-01
Fix security page

cllns:
1. Adds a `.html` file extension in the middle of the security file, which fixes its content-type, and automatically uses the right layout.
2. Switches the security contact to `admin@hanamirb.org`

Closes #502. Thanks again to @dgollahon for reporting this 🌸


AlfonsoUceda (approved):

dgollahon:
👍 

FYI, @cllns I think you may have meant to close #508 instead of #502.

PR #517, opened 2021-01-15
Fix incorrectly italicized title in v1.3.0 release notes

mjq:
This was my bad - when the deprecated method name was updated in 0f39c61f, Markdown started treating the underscores in `parsed_request_body` as italics.  Escaped the underscores. Tested this by running `bin/site build` locally and the section title shows up correctly. Sorry for the noise!


jodosha (approved):
@mjq Thanks again!

PR #522, opened 2021-05-04
Add "Announcing Hanami v2.0.0.alpha2" post

timriley:


jodosha (approved):
@timriley Thanks for making this to happen! 🌸

source/blog/2021-05-04-announcing-hanami-200alpha2.html.markdown
```diff
@@ -0,0 +1,376 @@
...
+```
+
+And this same view used within Hanami looks like this:
+
+```ruby
+class MyView < Hanami::View
+end
+```
+
+There’s almost no difference! Once you understand how to use `Hanami::View` once, you can use it everywhere. Even inside an Hanami app, where the app seamlessly integrates the views (in the case above, inferring the template name automatically, among other things), you can still access the full extent of the view configuration, allowing you to ”eject” from the configured defaults if you ever need.
```
solnic, line 319:
```suggestion
There’s almost no difference! Once you understand how to use `Hanami::View`, you can use it everywhere. Even inside an Hanami app, where the app seamlessly integrates the views (in the case above, inferring the template name automatically, among other things), you can still access the full extent of the view configuration, allowing you to ”eject” from the configured defaults if you ever need.
```

timriley: Tweak slightly to make it read better, thanks @solnic :)

PR #539, opened 2022-01-20
Fix top banner on tiny screens & update link

cllns:
This updates the banner link to just go to `/blog/` so that we don't have to update it every month. It was outdated anyway, so this is a more 'permanent' fix.

Before:
image

After:
image

It also decreases the height slightly on desktop, but it's still plenty big.


timriley:
Thanks @cllns, it will be nice not to have to think about this every month :)

PR #549, opened 2022-06-23
Fixes incorrect require in alpha8 release notes

ehannes:
I guess the Action base class should require the `hanami/application/action` and not `/view`, but I might be totally wrong though 😄


timriley:
Thank you! It's honestly nice to know that people are also reading these 😅

ehannes:
@timriley I'm reading everything I can get hold of right now 😄  We decided to start a production application in Hanami 2 quite recent. We believe in the project and besides the obvious "bleeding edge" problem we are very satisfied with Hanami 2. Thanks for all your time you are spending on the project! I wish you and the team all the best, continue your awesome work!

PR #550, opened 2022-07-19
Announcing Hanami v2.0.0.beta1

jodosha:


source/blog/2022-07-20-announcing-hanami-200beta1.html.markdown
```diff
@@ -0,0 +1,133 @@
...
+
+Let’s dissect a newly generated application (this one being named “Bookshelf”):
+
+- `app/` is home for your application code
+- `app/actions/` holds your actions and `app/action.rb`, with `Bookshelf::Action` as their superclass
+- `config/app.rb` defines your application (`Bookshelf::App`) and its configuration
+- `config/settings.rb` defines your app’s settings, such as the database URL
+- `config/routes.rb` defines your app’s HTTP routes
+- `lib/` is the home for any general purpose code used by your application code
+- `lib/bookshelf/types.rb` defines your custom Ruby data types (using the dry-types gem)
```
timriley, line 29: You know, I wonder if could also go without generating a types module for basic Hanami apps...

From the perspective of getting started, it'd be useful for type-checking your settings (which is optional), but there's nothing else about a newly-generated Hanami app that uses the types.

PR #555, opened 2022-09-15
Add note about Hanami 1.3 and Ruby 2

cllns:
We already do this on the Getting Started guide, but we've had requests for it on the homepage too.

https://github.com/hanami/guides/pull/91

https://github.com/hanami/hanami/issues/1108#issuecomment-1247657691

I couldn't get the server to work locally so let's see what the preview looks like :)


cllns:
Tweaked the styles too, to look better on mobile screens. Merged because it looked good

PR #571, opened 2023-03-04
Improve sizing and spacing of logos in trusted by section of homepage

ZimbiX:
The logos are a bit cramped currently, and the sizing and positioning of the Hippo and FlooringStores logos doesn't fit in. This change addresses this.

## Before - wide

image

## After - wide

image

## Before - narrow

image

## After - narrow

image

PR #573, opened 2023-03-04
Add title text to trusted by logos, including nationality

ZimbiX:
Depends on https://github.com/hanami/hanami.github.io/pull/571 - merge that first.

I was curious about the nationality of each company listed as using Hanami, so thought it would be useful to record that here.


ZimbiX:
Closing due to the disappointing removal of the logos section in https://github.com/hanami/hanami.github.io/pull/586

timriley:
@ZimbiX We're building a new site, and will consider bringing a logos section back if we can get a good slate of logos to share.

PR #586, opened 2024-05-06
remove_featured_section

krzykamil:
@timriley removing the section you mentioned, since it is too outdated to maintain and you want to overhaul the site (which I would gladly help with if needed)

image


timriley:
Thanks @krzykamil!

PR #587, opened 2024-06-26
Remove references to Gitter, and emphasize Discourse as preferred place

cllns:
Unfortunately the link here just goes to the Discourse forum if you're not logged in. It doesn't show any reference to the chat at all, it's even missing from the sidebar unfortunately. Is there any way to make the chat public @timriley?

Regardless, I wrote some copy that should be good enough regardless of whether it's visible or not. We can add some more text if it's not possible to make it public.

As always, open to feedback or changes to the copy.

### DIFF
--- a/source/_footer.erb
+++ b/source/_footer.erb
@@ -65,8 +65,3 @@
   ga('create', 'UA-47369640-1', 'hanamirb.org');
   ga('send', 'pageview');
 </script>
-
-<script>
-  ((window.gitter = {}).chat = {}).options = { room: 'hanami/chat' };
-</script>
-<script src="//sidecar.gitter.im/dist/sidecar.v1.js" async defer></script>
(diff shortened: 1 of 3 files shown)


timriley (approved):
This looks great, thanks for updating this, @cllns!

Unfortunately there's no way to make the available chat channels visible to non-signed-in users, [see here for details](https://meta.discourse.org/t/allow-visitors-and-search-engines-to-see-chat-channels/254267).

Maybe we could think about pinning some topics explaining the presence of the chat or something similar.

cllns:
I just added another commit to remove Reddit, StackOverflow and Awesome Hanami. The last one is cool but it's basically all Hanami 1 so not very relevant. It'd be nice to bring it back and updated it for Hanami 2 (and remove all Hanami 1 stuff), but not right now.

timriley (approved):
Agree with the removal of those things too, @cllns! Thank you!

PR #588, opened 2024-07-14
Add release announcement for 2.2.0.beta1

timriley:
[Readable](https://github.com/hanami/hanami.github.io/blob/hanami-220beta1/source/blog/2024-07-16-hanami-220beta1.html.markdown)


source/blog/2024-07-16-hanami-220beta1.html.markdown
```diff
@@ -0,0 +1,236 @@
...
+
+      def order_by_latest
+        order(self[:id].desc)
+      end
+    end
+  end
+end
+```
+
+Relations are the primary component of our app’s database layer. While we can interact with them directly across our app, it’s better to build a repo. With repos, we get to build our very own API to your persisted data, so we can better manage how it’s accessed across your app. We can build a post repo with `hanami generate repo postss`. Here we can define a method to return our latest posts:
```
alassek: *typo*: postss

timriley: Thanks, fixed!

PR #598, opened 2025-03-02
Fix the link to Getting Started Guide

y-yagi:
This is Hanami 2.2 release announcement. So guide for v2.2 is more appropriate I think.


runlevel5 (approved):

timriley:
Oh yes, that's very important, thank you for fixing!

PR #599, opened 2025-03-14
Add `rel="me"` attribute to Mastodon account link

jgarber623:
Thanks for considering this change!

Mastodon supports profile verification out-of-the-box:

https://joinmastodon.org/verification

This change adds the necessary `rel="me"` attribute to the link to Hanami's Mastodon profile in the website footer. Someone with admin access to the Mastodon account could then add a link to `https://hanamirb.org` by navigating to:

https://ruby.social/settings/profile

…and adding the link to the "Extra Fields" configuration.

Boom, verified links!

(also, it'd be super helpful to have a link to Hanami's homepage when viewing y'all's Mastodon profile)

### DIFF
--- a/source/_footer.erb
+++ b/source/_footer.erb
@@ -20,7 +20,7 @@
           <li><a href="/community#code-of-conduct">Code of Conduct</a></li>
           <li><a href="https://discourse.hanamirb.org" target="_blank">Forum</a></li>
           <li><a href="https://discord.gg/KFCxDmk3JQ" target="_blank">Chat</a></li>
-          <li><a href="https://ruby.social/@hanami" target="_blank">Mastodon</a></li>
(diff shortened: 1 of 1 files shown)


timriley:
Great idea, thanks @jgarber623!

timriley:
I've just added the website to https://ruby.social/@hanami. Hopefully the verified tick takes effect soon.

jgarber623:
Thanks for the review and merge, @timriley!

The Mastodon source code is _deep_ and there are a handful of actions that might kick off the profile verification process (the `VerifyLinkService` in the source). There's also quite a bit that could go sideways and we (as normal users) would have no idea how to debug. 🙃 

`VerifyLinkService ` is called by `VerifyAccountLinksWorker` which in turn is called from either `UpdateAccountService` or `ActivityPub::ProcessAccountService`. _Those_ services are called by too many different things to list. Phew.

